PR #45, opened 2020-06-01
Rework dataloader

### LINKED ISSUE #43: Rework of dataloader
It might be a good idea to embed the dataloader inside the model, because each dataloader is specific to a model anyways, this would also more closely follow the lightning guidelines

```python
class MantisRCNN(MantisModule):
    @staticmethod
    def dataloader(self, <pytorch dataloader kwargs>):
        # Do the specialization
        return dataloader
```

It would also be logical to bring `item2training_sample` inside the model
```python
class MantisRCNN(MantisModule):
    @staticmethod
    def item2training_sample(item):
        # convert item to training sample
        ...
```

lgvaz:
Closes #43 

Implemented dataloader on MantisRCNN...

### DIFF
--- a/mantisshrimp/models/mantis_rcnn.py
+++ b/mantisshrimp/models/mantis_rcnn.py
@@ -4,6 +4,7 @@ from ..imports import *
 from ..utils import *
 from ..core import *
 from .mantis_module import *
+from ..data import *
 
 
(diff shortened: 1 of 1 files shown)


lgvaz:
This was opened only for demonstrations purposes

PR #58, opened 2020-06-07
testing pull docs

lgvaz:


lgvaz:
Build does not fail on warnings, should we raise all warnings?

lgvaz:
Annotations are being added though...

Issue #61, opened 2020-06-09
Restructuring Model folder


oke-aditya:
Following discussion on Slack and on issue #60 
We come to a structure as follows

so we can have a structure something like this
-> backbones (use torchvision + custom)
-> layers (for some layers that you would use in the backbone)
-> models
-----> model_name_folder (e.g. fasterrcnn)
----------> model.py (take help of backbones here)
----------> dataloader.py (with minor edits for every mode)

model.py includes train_step, validation_step, test_step.

And train.py (code for training and inference from the model) Would be in the examples folder.

Inheriting from rcnn to faster rcnn is extra inter code coupling which we might avoid.
Let's have seperate structures for rcnn, fast rcnn and faster rcnn. It would make debugging easier as well.

Also, I will raise a PR for contributing.MD and FAQs.MD (will check how to make .rst)

lgvaz:
Could you explain what do you have in mind to be inside `dataloader.py`? Currently a dataloader is something as short as:
```python
    def dataloader(cls, dataset, **kwargs) -> DataLoader:
        def collate_fn(data):
            ts = [cls.build_training_sample(**o) for o in data]
            return list(zip(*ts))

        return DataLoader(dataset=dataset, collate_fn=collate_fn, **kwargs)
```

I think it makes much more sense to have this as an method of the class than unnecessary splitting in another file

lgvaz:
Also, as we discussed in slack, I still think `FasterRCNN` and `MaskRCNN` should inherit from a common parent class that implements `training_step` and `validation_step`, these methods are exactly the same for both models.

If you take a look on how torchvision implemented it, they have `MaskRCNN` that inherits from `FasterRCNN` that inherits from `GeneralizedRCNN`, I think we should draw inspiration from this.

I'm not saying to follow the exact same structure, I'm just saying we need a common `RCNN` parent

oke-aditya:
Dataloader can be method of the file, I think no need of different of file. 

But the inheriting from RCNN is something possible only in RCNN style models, would it be possible for SSD, YOLO etc ? It would lead to an inconsistent API.

This multiple inheritance makes it a bit hard to customize I felt.

lgvaz:
> But the inheriting from RCNN is something possible only in RCNN style models

That's exactly right! I believe this actually makes the code very clear instead of confusing, it wouldn't make sense to do `YOLO(RCNNModule)`, YOLO is not an RCNN extension!! hahaha

[...]

PR #64, opened 2020-06-09
Add backbones

oke-aditya:
- I have added the code to support most torchvision bacbones (except alexnet and Densenet)

I have unit tested the ` mantisshrimp/backbones/torchvision_backbones.py ` file.

How do I test the entire fasterRCNN code I am unsure. Please let me know so that I can test and create PRs next time.


oke-aditya:
Also, I guess there are some parameters with black formatting. 

I have done simply black filename.py I guess there is something more.

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,54 @@
...
+        else:
+            # Creates the custom backbone model trained on ImageNet.
+            try:
+                self.base_model = create_torchvision_backbone(
+                    backbone=self.backbone, is_pretrained=pretrained
+                )
+                self.ft_anchor_generator = AnchorGenerator(
+                    sizes=((32, 64, 128)), aspect_ratios=((0.5, 1.0, 2.0))
+                )
+                self.ft_mean = [0.485, 0.456, 0.406]  # imagenet mean and std
```
lgvaz, line 45: I don't think we need to specify the mean and std, FasterRCNN will default to that internally already

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,54 @@
...
+            self.m.roi_heads.box_predictor = FastRCNNPredictor(
+                in_features, self.n_class
+            )
+        else:
+            # Creates the custom backbone model trained on ImageNet.
+            try:
+                self.base_model = create_torchvision_backbone(
+                    backbone=self.backbone, is_pretrained=pretrained
+                )
+                self.ft_anchor_generator = AnchorGenerator(
```
lgvaz, line 42: Where is this being used?

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,54 @@
...
+                    sizes=((32, 64, 128)), aspect_ratios=((0.5, 1.0, 2.0))
+                )
+                self.ft_mean = [0.485, 0.456, 0.406]  # imagenet mean and std
+                self.ft_std = [0.229, 0.224, 0.225]
+                self.m = FasterRCNN(
+                    backbone=self.base_model,
+                    num_classes=self.n_class,
+                    image_mean=self.ft_mean,
+                    image_std=self.ft_std,
```
lgvaz, line 51: We should pass kwargs to `FasterRCNN` here, so the user can still specify parameters if we wants to

mantisshrimp/backbones/torchvision_backbones.py
```diff
@@ -0,0 +1,85 @@
+# This imports the torchivsion defined backbones
+__all__ = ["create_torchvision_backbone"]
+
+from ..imports import *
+
+
+def create_torchvision_backbone(backbone: str, is_pretrained: bool):
```
lgvaz, line 7: Can we change `is_pretrained` to `pretrained`?

I see your pointed of having `is_` and it does makes more sense, but I think it's better to follow torchvision convention here, it's what people are already used to

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,54 @@
...
-        in_features = self.m.roi_heads.box_predictor.cls_score.in_features
-        self.m.roi_heads.box_predictor = FastRCNNPredictor(in_features, self.n_class)
+        self.n_class, self.h, self.backbone, self.pretrained = (
+            n_class,
+            h,
+            backbone,
+            pretrained,
+        )
+
+        if self.backbone == "defualt":
```
lgvaz, line 29:
```suggestion
        if self.backbone == "default":
```

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
lgvaz: Same goes for the anchor generator, any reason you're explicit defining that instead of using torchvision defaults?

lgvaz:
Another tip, we can mention the issue here so it gets automatically closed when this merges =)

Closes #61

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya:
```
    def __init__(self, backbone, num_classes=None,
                 # transform parameters
                 min_size=800, max_size=1333,
                 image_mean=None, image_std=None,
                 # RPN parameters
                 rpn_anchor_generator=None, rpn_head=None,
                 rpn_pre_nms_top_n_train=2000, rpn_pre_nms_top_n_test=1000,
                 rpn_post_nms_top_n_train=2000, rpn_post_nms_top_n_test=1000,
                 rpn_nms_thresh=0.7,
                 rpn_fg_iou_thresh=0.7, rpn_bg_iou_thresh=0.3,
                 rpn_batch_size_per_image=256, rpn_positive_fraction=0.5,
                 # Box parameters
                 box_roi_pool=None, box_head=None, box_predictor=None,
                 box_score_thresh=0.05, box_nms_thresh=0.5, box_detections_per_img=100,
                 box_fg_iou_thresh=0.5, box_bg_iou_thresh=0.5,
                 box_batch_size_per_image=512, box_positive_fraction=0.25,
                 bbox_reg_weights=None):

        if not hasattr(backbone, "out_channels"):
            raise ValueError(
                "backbone should contain an attribute out_channels "
                "specifying the number of output channels (assumed to be the "
                "same for all the levels)")

        assert isinstance(rpn_anchor_generator, (AnchorGenerator, type(None)))
        assert isinstance(box_roi_pool, (MultiScaleRoIAlign, type(None)))

        if num_classes is not None:
            if box_predictor is not None:
                raise ValueError("num_classes should be None when box_predictor is specified")
        else:
            if box_predictor is None:
                raise ValueError("num_classes should not be None when box_predictor "
                                 "is not specified")

        out_channels = backbone.out_channels

        if rpn_anchor_generator is None:
            anchor_sizes = ((32,), (64,), (128,), (256,), (512,))
            aspect_ratios = ((0.5, 1.0, 2.0),) * len(anchor_sizes)
            rpn_anchor_generator = AnchorGenerator(
                anchor_sizes, aspect_ratios
            )
        if rpn_head is None:
            rpn_head = RPNHead(
                out_channels, rpn_anchor_generator.num_anchors_per_location()[0]
            )

        rpn_pre_nms_top_n = dict(training=rpn_pre_nms_top_n_train, testing=rpn_pre_nms_top_n_test)
        rpn_post_nms_top_n = dict(training=rpn_post_nms_top_n_train, testing=rpn_post_nms_top_n_test)

        rpn = RegionProposalNetwork(
            rpn_anchor_generator, rpn_head,
            rpn_fg_iou_thresh, rpn_bg_iou_thresh,
            rpn_batch_size_per_image, rpn_positive_fraction,
            rpn_pre_nms_top_n, rpn_post_nms_top_n, rpn_nms_thresh)

        if box_roi_pool is None:
            box_roi_pool = MultiScaleRoIAlign(
                featmap_names=['0', '1', '2', '3'],
                output_size=7,
                sampling_ratio=2)

        if box_head is None:
            resolution = box_roi_pool.output_size[0]
            representation_size = 1024
            box_head = TwoMLPHead(
                out_channels * resolution ** 2,
                representation_size)

        if box_predictor is None:
            representation_size = 1024
            box_predictor = FastRCNNPredictor(
                representation_size,
                num_classes)

        roi_heads = RoIHeads(
            # Box
            box_roi_pool, box_head, box_predictor,
            box_fg_iou_thresh, box_bg_iou_thresh,
            box_batch_size_per_image, box_positive_fraction,
            bbox_reg_weights,
            box_score_thresh, box_nms_thresh, box_detections_per_img)

        if image_mean is None:
            image_mean = [0.485, 0.456, 0.406]
        if image_std is None:
            image_std = [0.229, 0.224, 0.225]
        transform = GeneralizedRCNNTransform(min_size, max_size, image_mean, image_st
```

oke-aditya: So by default they are getting assigned.
The reason I passed was beacause of image_size < Anchor Size and user needs to take care of that. But I guess adding **kwargs user will be able to edit this.

oke-aditya: I am removing them from coed and adding **kwargs.

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya: Resolved from previous conversation. I am not putting them altogether allowing **kwargs.

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya: This is gonna be done.

mantisshrimp/backbones/torchvision_backbones.py
oke-aditya: Changed to follow convention

oke-aditya:
> 
> 
> Thanks for the PR!
> 
> There are some suggestions I would like to give before merging
> 
> Currently, how can a user specify a custom backbone that is not defined in `create_torchvision_backbone`? The way is implemented right now he would need to change the source code
> 
> What do you think of having something like
> 
> ```python
> backbone = get_torchvision_backbone('mobile_net')
> model = MantisFasterRCNN(backbone=backbone, ...)
> ```
> 
> Also, keep in mind that we want to support pytorch hub in the future, how can we write something that we can easily integrate with that?

For Pytorch hub, we could simply take CNN features from there. Maybe create a file in backbone folder as ```get_hub_backbones``` and make use of them here.

For user specified backbones, I'm unsure, user would have to write their CNN architecture and get features from their CNN architecture. This needs to be passed to our our code.

oke-aditya:
I have made changes, to solve other stuff but we would need to think to accommodate those two as well.

oke-aditya:
Also, I just checked the research paper of FasterRCNN. Originally the backbones were trained with ImageNet and the other layers were fine-tuned on COCO.

So the torchvision layers with pretrained ImageNet weights should be fine.

oke-aditya:
As suggested in the paper Anchors are not optimized in FasterRCNN it should be taken care as per the dataset needs. I guess users are aware of this.

lgvaz:
This is kinda of what I had in mind:
```python
class MantisFasterRCNN(MantisRCNN):
    @delegates(FasterRCNN.__init__)
    def __init__(self, n_class, backbone=None, metrics=None, **kwargs):
        super().__init__(metrics=metrics)
        self.n_class = n_class

        if backbone is None:
            # Creates the default fasterrcnn as given in pytorch. Trained on COCO dataset
            self.m = fasterrcnn_resnet50_fpn(pretrained=True, **kwargs)
            in_features = self.m.roi_heads.box_predictor.cls_score.in_features
            self.m.roi_heads.box_predictor = FastRCNNPredictor(in_features, n_class)
        else:
            self.m = FasterRCNN(backbone=backbone, num_classes=n_class, **kwargs)
```

lgvaz:
Actually, looking at torchvisions source code, I see they already have a custom function for getting backbones [resnet_fpn_backbone](https://github.com/pytorch/vision/blob/master/torchvision/models/detection/backbone_utils.py#L44), it also add FPN on top which is nice.

Any disadvantages of using this instead of `create_torchvision_backbone`?

oke-aditya:
It has support for all ResNet models. Suppose we want to extend for other models such as VGG 16 (originally in paper) and mobile net. Or in future from hub. We would need a generalized function. Hence I thought of this.

oke-aditya:
https://pytorch.org/docs/stable/_modules/torchvision/models/detection/faster_rcnn.html#fasterrcnn_resnet50_fpn

As suggested we could pass the backbones as well, but it would create without fpn of resnet. I guess we can have an extra user argument to specify if they are using resnet, they need fpn for resnet or not.

For Backbones like vgg, mobilenet, we wouldn't be providing fpn as of now. Also if user wants to pass their own CNN as backbone again we won't provide fpn.

oke-aditya:
FPN is completely optionally, Faster RCNN can work without it. So let us give user the choice, it may perform better or worse according to user needs.

oke-aditya:
Added code for optional fpn and allowing to instatiate a CNN model as backbone. 
Need to test and fix the things.
I am writiting pytest for FasterRCNN.

lgvaz:
> FPN is completely optionally, Faster RCNN can work without it. So let us give users the choice.

I completely agree that the users should have the choice. But keep in mind that our defaults should always be on the best practices 😁

You probably already did this, I'm commenting before reviewing the code, I thought it's important to make this clear

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,89 @@
...
+        **kwargs,
+    ):
         super().__init__(metrics=metrics)
-        self.n_class, self.h, self.pretrained = n_class, h, pretrained
-        self.m = fasterrcnn_resnet50_fpn(pretrained=self.pretrained, **kwargs)
-        in_features = self.m.roi_heads.box_predictor.cls_score.in_features
-        self.m.roi_heads.box_predictor = FastRCNNPredictor(in_features, self.n_class)
+        self.n_class, self.h, self.backbone, self.pretrained, self.fpn = (
+            n_class,
+            h,
```
lgvaz, line 25: As you commented previously, h is not being used right? You can remove it

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,89 @@
...
+            backbone,
+            pretrained,
+            fpn,
+        )
+
+        supported_resnet_models = [
```
lgvaz, line 31: I think it might be a good idea to implement this as a class attribute, and maybe rename it to `supported_resnet_fpn_models`.

The user can then do `MantisFasterRCNN.supported_resnet_fpn_models` to know what options he can pass.

Also, you can add the supported models to the docstring 😊

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,89 @@
...
+                    )
+                    self.m = FasterRCNN(self.backbone, self.n_class, **kwargs)
+                else:
+                    raise NotImplementedError(
+                        "FPN for non resnets is not supported yet"
+                    )
+
+            else:
+                # This does not create fpn backbone, it is supported for all models
+                self.base_model = create_torchvision_backbone(
```
lgvaz, line 74: base_model is just an implementation detail, we don't need to save it in `self`

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,89 @@
...
+                    backbone=backbone, pretrained=self.pretrained
+                )
+                self.m = FasterRCNN(
+                    backbone=self.base_model, num_classes=self.n_class, **kwargs
+                )
+
+        elif isinstance(self.backbone, torch.nn.modules.container.Sequential):
```
lgvaz, line 81: Just a curious question, why check for `Sequential` instead of `nn.Module`?

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,89 @@
...
+            )
+
+        elif isinstance(self.backbone, str):
+            # Giving string as a backbone, which is either supported resnet or backbone
+            if self.fpn is True:
+                # Creates a torchvision resnet model with fpn added
+                # Will need to add support for other models with fpn as well
+                # Passing pretrained True will initiate backbone which was trained on ImageNet
+                if self.backbone in supported_resnet_models:
+                    self.m = resnet_fpn_backbone(
```
lgvaz, line 63:
```suggestion
                    self.backbone = resnet_fpn_backbone(
```

lgvaz: Actually it's better just to call it backbone (without self), and we let self.backbone be whatever was the original argument the user passed (can be useful for logging in the future)

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
```diff
@@ -4,16 +4,89 @@
...
+        fpn=True,
+        metrics=None,
+        **kwargs,
+    ):
         super().__init__(metrics=metrics)
-        self.n_class, self.h, self.pretrained = n_class, h, pretrained
-        self.m = fasterrcnn_resnet50_fpn(pretrained=self.pretrained, **kwargs)
-        in_features = self.m.roi_heads.box_predictor.cls_score.in_features
-        self.m.roi_heads.box_predictor = FastRCNNPredictor(in_features, self.n_class)
+        self.n_class, self.h, self.backbone, self.pretrained, self.fpn = (
```
lgvaz, line 23: Can we instead have one definition per line?

I also used to do it like this, but got convinced that doing one per line is a more readable way (we can see the formatter already breaks into multiple lines anyways, so no advantage in doing it like this)

tests/backbones/test_backbones.py
```diff
@@ -0,0 +1,27 @@
+from mantisshrimp.backbones import *
+import pytest
+import torch
+
+
+def test_torchvision_backbones():
+    supported_backbones = [
```
lgvaz, line 7: If we define the supported backbones as a class method, we don't have to redefine them here again 😃

tests/models/test_fastercnn_creation.py
```diff
@@ -0,0 +1,81 @@
+import pytest
+from mantisshrimp.models.mantis_rcnn import *
+import torch
+
+
+def test_fastercnn():
+    # We need to instantiate with all possible combinations
+    # Taken directly from mantis_faster_rcnn
+    supported_resnet_models = [
```
lgvaz, line 9: Yeah, again, let's figure out a way of defining those in the model

tests/models/test_fastercnn_creation.py
```diff
@@ -0,0 +1,81 @@
...
+            fpn=is_fpn,
+        )
+        assert isinstance(
+            model, mantisshrimp.models.mantis_faster_rcnn.MantisFasterRCNN
+        )
+
+    # Check for simple CNN that can be passed to make backbone
+    # Will think of an example and add soon
+
+    # Stuff to be added, running these models on a fake image data.
```
lgvaz, line 81: Perfect!

mantisshrimp/backbones/create_fpn_backbone.py
```diff
@@ -0,0 +1,107 @@
+from collections import OrderedDict
```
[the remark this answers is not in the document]
oke-aditya: Yes, this entire file is used no where so just deleting it completely. Again a liability for us if torchvision maintains it.

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya: Removed h completely from file. Even in init. I see no initialization internally as well.

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya: Added docstring 👍 A complete docstring for all params.

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya: Removed self

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya: Nothing specific as such, Since other torchvision backbones were of this class,
```
isinstance(ft_backbone, torch.nn.Module)
True
```
Replacing it to this. It works either ways.

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya: Reformatted, one definition per line now.

tests/backbones/test_backbones.py
oke-aditya: Defined as a class method with self inside mantis_faster_rcnn

tests/models/test_fastercnn_creation.py
oke-aditya: Defined as a class method with self inside mantis_faster_rcnn

tests/models/test_fastercnn_creation.py
oke-aditya: Also, I fixed a typo that caused the previous test to fail.

mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
oke-aditya: Kept self.backbone wherever user needed.
Used word backbone wherever backbone was created
Removed base model and renamed to backbone for consistency.

oke-aditya:
Yay ! All the models are instantiating, all tests are fixed now. 
Only thing left is running the Faster RCNN on a dummy image

oke-aditya:
So defualt RCNN with resnet 50 fpn is working fine, Some other isn't working. One reason maybe the image size is < Conv feature maps.

oke-aditya:
```
Testing backbone = resnet101
```
It is throwing error hmmm
I don't get the point though, other backbones with resnet fpn 18, 35, 50 worked. This didn't

oke-aditya:
Trying without Bigger Resnets like resnet 101 and 151 in all cases

oke-aditya:
Trying other backbone first, rereading images to fix it up.

oke-aditya:
Tests passing for all except 
```
# "resnet101",
        # "resnet152",
        # "resnext50_32x4d",
        # "resnext101_32x8d",
        # "wide_resnet50_2",
        # "wide_resnet101_2",
```

oke-aditya:
I guess we can go ahead with this for now. 
4 stuff remain to be solved.
- Passing tests for remaining ResNet models,
- Tests for custom CNN backbones
- Adding FPN for any arbitary backbone (or atleast those in torchvision).
- Also support any reamaining torchvision backbones (guess densenet and alexnet, inception resnet).

oke-aditya:
It is unable to install dependencies. Strange, this time.

oke-aditya:
Strange, absolutely strange, Custom RCNN tests are passing in all three though, Resnet tests are still failing. This time only in 3.8.

oke-aditya:
```
tests/models/test_fastercnn_custom_cnn.py .                              [ 30%]
tests/models/test_fastercnn_nonresnets.py .                              [ 33%]
tests/models/test_fastercnn_resnets.py F                                 [ 35%]
```

Non resnets are fine. Resnets work half well and half don't. I have checked it locally it runs fine,

oke-aditya:
I have made the tests ultra granular now, non resnets non fpns, resnets not fpns, resnets fpns.
we do not support non resnets fpns as of now. 

But this is best testing I can try

oke-aditya:
- The issue maybe with loading the state dict to pretrained stuff. This causes github cache I guess. 
- This time trying without True pretrained

oke-aditya:
Absolutely fine 😀

oke-aditya:
https://github.com/pytorch/vision/issues/993

This is the problem. 

Torchvision does not support this maybe.
But its opposite to what is given in the docs 

https://github.com/pytorch/vision/blob/c2e8a00885e68ae1200eb6440f540e181d9125de/torchvision/models/detection/backbone_utils.py#L44

lgvaz:
Does this covers majority of the uses cases?
```python
@pytest.mark.slow
@pytest.mark.parametrize("pretrained", [False, True])
@pytest.mark.parametrize(
    "backbone, fpn",
    [
        ("mobilenet", False),
        ("vgg11", False),
        ("vgg13", False),
        ("vgg16", False),
        ("vgg19", False),
        ("resnet18", False),
        ("resnet34", False),
        ("resnet50", False),
        ("resnet18", True),
        ("resnet34", True),
        ("resnet50", True),
        # these models are too big for github runners
        # "resnet101",
        # "resnet152",
        # "resnext101_32x8d",
    ],
)
def test_faster_rcnn_nonfpn_backbones(image, backbone, fpn, pretrained):
    backbone = MantisFasterRCNN.get_backbone_by_name(
        name=backbone, fpn=fpn, pretrained=pretrained
    )
    model = MantisFasterRCNN(n_class=3, backbone=backbone)
    model.eval()
    pred = model(image)
    assert isinstance(pred, list)
    assert set(["boxes", "labels", "scores"]) == set(pred[0].keys())
```

oke-aditya:
Yes it does. I will do the testing for other cases as well.
We need to document this more carefully than relying on pytest.

lgvaz:
The problems on the tests were being caused when the model was on `.eval` mode and did not found any boxes in the image.

Fixed it by testing on training mode, with the additional benefit that it also tests `build_training_sample` now (indirectly, but okay)

lgvaz:
Please review the proposed changes, if you approve and the tests pass, I'll go ahead and merge this.

Many thanks!!!

oke-aditya:
Great. All tests are passing now. 👍 Can't wait for this feature to be merged.

PR #65, opened 2020-06-10
Add faqs

oke-aditya:
Adding FAQs rst.


oke-aditya:
I wanted to add citing Mantisshrimp. Maybe possible sometime later.

Issue #93, opened 2020-06-18
UnicodeDecodeError when installing


tazu786:
## 🐛 Bug
**Describe the bug**
Error when I install through pip.

**To Reproduce**
1. run `pip install git+git://github.com/lgvaz/mantisshrimp.git`

>... 
"c:\users\...\appdata\local\programs\python\python37\lib\encodings\cp1252.py", line 23, in decode
        return codecs.charmap_decode(input,self.errors,decoding_table)[0]
    UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 554: character maps to <undefined>
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.

**Expected behavior**
Installation

**Desktop (please complete the following information):**
 - OS: [win 10]

**Additional context**

oke-aditya:
Thank you for the bug report. I am too able to reproduce the bug. 

OS windows 10 pro 64bit
Python version 3.6

```
PS C:\Users\hp> pip uninstall manstisshrimp
WARNING: Skipping manstisshrimp as it is not installed.
PS C:\Users\hp> pip install git+git://github.com/lgvaz/mantisshrimp.git
Collecting git+git://github.com/lgvaz/mantisshrimp.git
  Cloning git://github.com/lgvaz/mantisshrimp.git to c:\users\hp\appdata\local\temp\pip-req-build-piohh5yv
  Running command git clone -q git://github.com/lgvaz/mantisshrimp.git 'C:\Users\hp\AppData\Local\Temp\pip-req-build-piohh5yv'
  Running command git submodule update --init --recursive -q
    ERROR: Command errored out with exit status 1:
     command: 'c:\program files\python36\python.exe' -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\hp\\AppData\\Local\\Temp\\pip-req-build-piohh5yv\\setup.py'"'"'; __file__='"'"'C:\\Users\\hp\\AppData\\Local\\Temp\\pip-req-build-piohh5yv\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base 'C:\Users\hp\AppData\Local\Temp\pip-pip-egg-info-s1vdaybl'
[...]

PR #95, opened 2020-06-18
Fix setup.py installation issue on Windows.

oke-aditya:
Fixes Issue #93.


oke-aditya:
I could install manually after adding this fix and running `pip install .`

Issue #94, opened 2020-06-18
Documentation colab badge does not render


lgvaz:
## 🐛 Bug
**Describe the bug**
Colab badge does not render properly on documentation. But it does locally.

**To Reproduce**
Take a look [here](https://lgvaz.github.io/mantisshrimp/tutorials/hub_detr_finetune_pennfundan.html)

**Expected behavior**
Show the colab badge

**Screenshots**
![Screenshot 2020-06-18 at 3 03 46 PM](https://user-images.githubusercontent.com/16520544/85056036-eb496580-b174-11ea-89f8-6e57f731b5fd.png)

oke-aditya:
Have a look [here](https://github.com/weiji14/deepbedmap/issues/61)

oke-aditya:
As well as [here](https://colab.research.google.com/github/googlecolab/colabtools/blob/master/notebooks/colab-github-demo.ipynb#scrollTo=8QAWNjizy_3O)

lgvaz:
The badge works correctly when inside a notebook, the problem is when it gets converted to an html page (via sphinx).

For some reason only the badges show but the link isn't there

oke-aditya:
Closed with the new documention #197 Closes this.

PR #96, opened 2020-06-18
Added colab badge.

paras-jain:
Taking a cue from [here](https://github.com/stellargraph/stellargraph/pull/1398). If this renders well on the docsite, will add badges to other tutorial notebooks as well. Addressing #94 .


lgvaz:
Thanks for the PR!

Last time I tried this it showed the image but it was not clickable, let's see what happens

lgvaz:
So yeah, I rolled back on using just a link again. A using an html ref correctly displays the image but it's not clickable

paras-jain:
Yeah just checked for some reason it's not clickable 🤔. Yes we should stick with the link, it serves the purpose atleast. Keep the issue open though.

Issue #99, opened 2020-06-19
Deployment


lgvaz:
## 🚀 Feature
Deployment, the thing we as ML devs always keep pushing forward can't be pushed forward no more.

Let's first discuss ideas and work on minimal examples before jumping head first on this one. I myself don't really know how deployment works, so I'll be starting with very simple apps and I will be sharing my journey here.

This is going to be a hard one, let's not do this alone, collective brain power is what we need right now, so I'll tag everyone 😅

@oke-aditya @paras-jain @ai-fast-track

lgvaz:
As we previously discussed, it's a good a idea to start with streamlit, so I'll do that

oke-aditya:
- Add docker support.
- Publish to PyPi experimental branch.
- Add support for Amazon Container Registry and Google Container Registry.
- Try with Kubernetes as well.
- Run with KubeFlow and AirFlow.

ai-fast-track:
Start small and build up on that. The way I see it:
1- Use streamlit
2- Deploy locally with a toy example (MantisFasterRCNN + Toy Dataset)
3- Use docker locally
4- Deploy docker container on the cloud:
5- Rinse and repeat

oke-aditya:
#97 should be the first big step.

Issue #97, opened 2020-06-19
Add Docker support with GPU


oke-aditya:
## 🚀 Feature
 
Add support for containerizing mantisshrimp.
We should create a Dockerfile, and a .dockerignore file

lgvaz:
That's a great idea, and it should make it very easy to install the lib, this would even solve the problem of pycocotools in windows right?

oke-aditya:
Yes, it will solve Pycocotools problem. Can you add CI for docker build on Github. So that we can check if container builds with our dockerfile.

lgvaz:
Can you make a PR for that?😄

oke-aditya:
I will make PR for the docker build yaml file. But you need to add the secrets.
Check [here](https://medium.com/better-programming/github-actions-for-docker-eaf22bbcc879)

oke-aditya:
Not yet solved with #104 . Will add a direct GPU supported NVIDIA container runtime docker itself. 

It will solve all the problems and will not create a multiple dockerfile issue.

ai-fast-track:
@oke-aditya, @lgvaz , Paperspace provides GPU-enabled container that are pre-loaded with both fastai2 library and the latest fastai course (V4). Check out the references here below, they might be adapted to mantisshrimp. 

https://github.com/Paperspace/fastai-docker/tree/master/fastai-v4
https://github.com/Paperspace/fastai-docker/blob/master/fastai-v4/Dockerfile
https://github.com/NVIDIA/nvidia-docker

oke-aditya:
Yes gonna adapt them 👍 We will have a docker file very similar to them.

PR #100, opened 2020-06-19
Add docker support

oke-aditya:
To add docker files. First, we need a CI-CD action to deploy it to docker hub.
The first step to #97 and #99.

@lgvaz you need to set up secrets as username and password for docker hub in your GitHub secrets action.

As described [here](https://medium.com/better-programming/github-actions-for-docker-eaf22bbcc879)


lgvaz:
Will this work without a Dockerfile?

oke-aditya:
Okay, I'm building locally. Once done will add PR for that too.

lgvaz:
Just keep pushing to this PR =)

oke-aditya:
Added minimal docker file. It should hopefully.

oke-aditya:
Everything works locally now. @lgvaz setting up the Secrets with GitHub should and CI action should push this to docker hub as well. 

As we update the master new container will be created and pushed again to docker container hub.

oke-aditya:
Yes, we can provide docker images with CUDA support as well as with No CUDA support (I guess this is useful for stuff like Raspberry Pi kinda devices). We need to tag them differently and need different Dockerfiles for both. Something which TensorFlow does [like this](https://hub.docker.com/r/tensorflow/tensorflow/tags). Currently, Dockerfile is built with torch GPU.

oke-aditya:
It is done as in [here](https://docs.docker.com/config/containers/resource_constraints/). The current image is built for GPU. As it uses PyTorch GPU. User needs to follow these instructions to run NVIDIA drivers within their container.

oke-aditya:
What best we can do is provide containers for both CPU and GPU separately. People need to setup their GPU using nvidia-docker and just run the container which will get created with this PR.

lgvaz:
Alright then, let's merge this one and we keep adding more support as it needs

PR #101, opened 2020-06-20
Add minimal docker

oke-aditya:
Added support for minimal docker cpu and gpu


oke-aditya:
It changes the CI for repo as well as adds seperate docker files. Dockerfile can be very much better, but for now lets get a MVP.

oke-aditya:
Correct. The GPU one is not installed using NVIDIA Cuda. We should do something like 
[this](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/docker/Dockerfile). 
With the current file, we need the user to install NVIDIA drivers and then run this. So that it runs on GPU. 

Let us build the minimal file first. We need to do lot of improvements on it. But will improvize one by one.

lgvaz:
> With the current file, we need the user to install NVIDIA drivers and then run this. So that it runs on GPU.

It's a bit weird to install NVIDIA drivers and then run this, because normally we directly pull from the image, which has an ubuntu base...

How would the user go to install nvidia and then pulling the image?

oke-aditya:
I will add support for nvidia-cuda docker container in an update (very shortly) for now this build is just to ensure everything is set up fine.

[This describes both the methods.](https://stackoverflow.com/questions/25185405/using-gpu-from-a-docker-container)

oke-aditya:
Current build is an Ubuntu Image with Packages installed in it. In a new PR will shift to nividia docker container.

PR #105, opened 2020-06-21
Adds fastai support

lgvaz:
This is going to be quite a big refactor, with this PR we're **truly** going to start support multiple engines for training.

This is not only about adding fastai support, but also for setting up the structure to support any other library

Start by taking a look at `examples/maskrcnn_fastai.py` and give your opinions on how the workflow currently looks like. Is it easy to understand? Are those steps intuitive? (ignore `examples_maskrcnn_lightning.py` for now, I'll refactor that)

Currently the file is a mess I know, but try to ignore that and focus on the steps taken, just by looking at it, do you have an idea how to do this with wheat for example?

The structure I end up going is `mantisshrimp/engines/*`, and this we'll keep adding (Lightning, Catalyst, etc..). My original though was to have those in hub, but I believe hub is starting to become something else, so it makes more sense to by in a separate folder.

After being comfortable with the example, take a look at the files under `engines/fastai` and tell if they make sense to you, I quite heavily changed my coding style in the past months, coming from "cool and compact" to "simple and stupid", so I hope those files are easy to understand, even though the logic behind it is quite hard.

You can safely **ignore the files in the other folders**, they're dirty as of now and I'm just experimenting.

I would like everyone to review and share your thoughts on this one, so I'm tagging you =)  
@oke-aditya @paras-jain @ai-fast-track 

---
Fun note: Defining functions and classes inside functions was a big no-no for me some months ago, but now you can see I do it very frequently, it really makes clear that the function/class was defined only to be used inside that function and no where else, with the added benefit that you don't have to keep jumping around in code.


lgvaz:
This PR is not complete, I'll soon be adding `engines/lightning` and then we'll see if this truly works out.

I just wanted to share this as early as possible so we start discussing it already

lgvaz:
I would suggest to checkout this branch and try running `maskrcnn_fastai`, maybe put it on a notebook or a interactive environment and try playing with the options

lgvaz:
Alright, so here is the important stuff.

## fastai
```python
learn = rcnn_learner(
    dls=[train_dataloader, valid_dataloader], model=model, metrics=[metric]
)

learn.fine_tune(3, lr=2e-4)
```

## lightning
```python
class LightModel(LightningModule):
    def __init__(self, model: MantisRCNN):
        super().__init__()
        self.model = model

    def training_step(self, batch, batch_idx):
        xb, yb = batch
        preds = self(xb, yb)
        loss = self.model.loss(preds, yb)
        log = {"train/loss": loss}
        return {"loss": loss, "log": log}

    def validation_step(self, batch, batch_idx):
        xb, yb = batch
        with torch.no_grad():
            self.train()
            preds = self(xb, yb)
            loss = self.model.loss(preds, yb)
            self.eval()
            # preds = self(xb)
        return {"valid/loss": loss}

    def validation_epoch_end(self, outs):
        log = {k: torch.stack(v).mean() for k, v in mergeds(outs).items()}
        return {"val_loss": log["valid/loss"], "log": log}

    def forward(self, *args, **kwargs):
        return self.model(*args, **kwargs)

    def configure_optimizers(self):
        opt = SGD(self.parameters(), 2e-4, momentum=0.9)
        return opt


light_model = LightModel(model)
trainer = Trainer(max_epochs=3, gpus=1)
trainer.fit(light_model, train_dataloader, valid_dataloader)
```

---

The difference is quite huge I know...

I thought about creating a `LightningRCNN` class or something like that to reduce the boilerplate, but  it would involve some assumptions on how the code should work. The idea of supporting multiple frameworks is so that people don't have to learn a new one, I believe lightning users are comfortable writing more code, and the proposed solution is pure lightning.

Also, this way, if you're doing something standard you use fastai. If you want to get fancy you can use lightning, I think we get the best of both worlds =)

What do you all think?

ai-fast-track:
I think we should keep the PL code as it is and not substituted by ours. PL users will recognize that code better than our custom adding. It also fairly compares the 2 libraries/frameworks.

lgvaz:
What I don't like about that is that every time you want to use `RCNNs` with lightning you will have to write all of that over and over again...

And in the snippet I gave, metrics are not even being handled.

But again, lightning people might be used and want exactly that...

lgvaz:
Second approach for lightning is to have something like this:

## Base class for all models

This handles all common functionality that all models should have. As of now, only metrics.
```python
class LightningModelAdapter(LightningModule, ABC):
    def __init__(self, metrics: List[Metric]):
        self.metrics = metrics
        self.reset_metrics()

    def reset_metrics(self):
        for metric in self.metrics:
            metric.reset()

    def accumulate_metrics(self, xb, yb, preds):
        for metric in self.metrics:
            metric.accumulate(xb, yb, preds)

    def finalize_metrics(self,):
        logs = {}
        for metric in self.metrics:
            value = metric.finalize()
            logs[metric.name] = value
        return logs
```

## Specific models 
This handles common functionality of a specific model:
```python
class RCNNLightningAdapter(LightningModelAdapter, ABC):
    def __init__(self, model: MantisRCNN, metrics: List[Metric]):
        super().__init__(metrics=metrics)
        self.model = model

    def forward(self, *args, **kwargs):
        return self.model(*args, **kwargs)

    def training_step(self, batch, batch_idx):
        xb, yb = batch
        preds = self(xb, yb)
        loss = self.model.loss(preds, yb)
        log = {"train/loss": loss}
        return {"loss": loss, "log": log}

    def validation_step(self, batch, batch_idx):
        xb, yb = batch
        with torch.no_grad():
            self.train()
            preds = self(xb, yb)
            loss = self.model.loss(preds, yb)
            self.eval()
            preds = self(xb)
            self.accumulate_metrics(xb, yb, preds)
        return {"valid/loss": loss}

    def validation_epoch_end(self, outs):
        loss_log = {k: torch.stack(v).mean() for k, v in mergeds(outs).items()}
        metrics_log = self.finalize_metrics()
        log = {**loss_log, **metrics_log}
        return {"val_loss": log["valid/loss"], "log": log}
```

## User model
This is what the user now has to define, just the remaining pieces for the model to work.
```python
class MyModel(RCNNLightningAdapter):
    def configure_optimizers(self):
        opt = SGD(self.parameters(), 2e-4, momentum=0.9)
        return opt
```

If the user wants to change behaviour, he can `super` the methods or reimplement them.

lgvaz:
@ai-fast-track , @oke-aditya , what do you think of the new lightning proposal?

oke-aditya:
This approach is much better.
It will allow easy refactoring of code as well.

ai-fast-track:
Using those adapters, PL might decide to go this route to have a concise DRY API!

lgvaz:
@ai-fast-track I was not able to completely understand your comment 😅

It means you like this new approach?

ai-fast-track:
What I meant is your approach is cleaner and it eliminates boilerplate.

lgvaz:
Alright! Getting there!!

I think the API is reasonably good at this point. All files are clean (apart from `example/maskrcnn_fastai.py`) and almost ready to merge.

I also updated the wheat notebook, can you all check it and see if it makes sense?

I still have not run the full experiment with both frameworks, it would be interesting to see the difference in performance between fastai and lightning (I'm expecting fastai to have a better result because it's using freeze/unfreezing and then differential learning rates, but the optimizers currently used are also different).

Would be nice to see a colab notebook comparing both =)

Issue #98, opened 2020-06-19
Pycoco tools do not install / run on Windows


oke-aditya:
## 🐛 Bug
**Describe the bug**

Simple this isn't our side issue though but if people come here they should know about this.

oke-aditya:
Pycoco tools do not run on windows. We are looking to use other metrics.
[Issue with Pycoco tools](https://github.com/cocodataset/cocoapi/issues/9#issuecomment-217339740)

For now issue can be solved with docker container #97.
Or use [this ](https://github.com/philferriere/cocoapi)

oke-aditya:
- For now, Windows users can use this as a solution. We have tested this. But do not officially support or nor maintain the repo to support this. It is just a workaround which has worked well.

You can find additional details in README of [this](https://github.com/philferriere/cocoapi) coco api repo.
- In short

1. Install any dependency if missing like cython, etc.
2.  Install Visual C++ 14/2015 build tools
3.  `pip install git+https://github.com/philferriere/cocoapi.git#subdirectory=PythonAPI`

oke-aditya:
Another solution is to use Docker with Linux backend containers and run mantisshrimp docker image in it.

ai-fast-track:
In Windows, Docker does not support GPU for now (GPU support will be in WSL2)

oke-aditya:
- WSL2 is already launched with Windows 10 May 2020 update. I am using WSL 2 as docker linux backend.

ai-fast-track:
It wasn't publicly released: You have to enroll in their beta testing group. The installation seems to be quiet a challenge based on this fastai thread: [Platform: Windows 10 using WSL2 w/GPU](https://forums.fast.ai/t/platform-windows-10-using-wsl2-w-gpu/73521)

oke-aditya:
It is quite simple. The above can be a good reference to how to configure this.
1. Have Windows 10 May 2020 Update (minimum)
2. Have your NVIDIA GPU with proper CUDA installed. This the user can refer to [NVIDA](https://developer.nvidia.com/cuda/wsl) 
3. Install WSL2. Info [here](https://docs.microsoft.com/en-us/windows/wsl/install-win10)
4. Setup WSL2. Again a brief commands. Given in the above page.
5. Install Ubuntu from Windows store. Should be a few clicks.
6. Install Docker it will give you option of using WSL2 backend. Or refer Docker docs to set it up. Fairly simple.
[...]

PR #106, opened 2020-06-22
Added GPU Docker support

oke-aditya:
- Uses Docker best practices.
- I have checked it builds locally.
- Uses NVIDIA Docker GPU runtime. So torch accelerates with GPU.

Docker action is too changed so that it build with this.


oke-aditya:
Will fix #97 . Helps in #99 . Helps in #98

oke-aditya:
Yes. This single file does all the job of running torch and mantisshrimp on GPU, user just needs to have their GPU initialized properly.

ai-fast-track:
I find it a bit confusing to have these 2 similar  issues (#97 and #106) running in parallel. Can we merge them?

oke-aditya:
Clarification: -
This is CPU + GPU build. It works fine on CPU only devices as well.

Below are my screenshots when I tested it on CPU only device.
image

image

It will work fine on GPU device as well. It has torch-gpu installed.

oke-aditya:
> 
> 
> Clarification: -
> This is CPU + GPU build. It works fine on CPU only devices as well.
> 
> Below are my screenshots when I tested it on CPU only device.
> image
> 
> image
> 
> It will work fine on GPU device as well. It has torch-gpu installed.

#97 is the only issue and #106 is the PR as solution to it. Which is this

oke-aditya:
Fixed all the stuff. It is working absolutely fine. Able to import in both cases without absolute imports and with it too.

PR #109, opened 2020-06-22
replaces relative imports with absolute

### LINKED ISSUE #46: Start using absolute imports
Absolute imports are easier to refactor (automatically refactor when moving packages) and offer no drawbacks

lgvaz:
closes #46 

@oke-aditya can you verify if this branch now works with docker?


oke-aditya:
I checked by building a docker image and running it. It works fine.

PR #114, opened 2020-06-23
Fixes the docs

oke-aditya:
Due to misplace of ` ## ` it did not render properly.

This should fix it.


lgvaz:
Can I also use this opportunity to ask for you to remove all the output from `Trainer.fit`?

Just comment the cell, run it, and uncomment again 😅

Thanks!

oke-aditya:
K. Done that too.

Issue #75, opened 2020-06-14
Integrate Pytorch Image Models as Hub Module


oke-aditya:
## 🚀 Feature
Many SOTA implementations in Pytorch are available in this well maintained repo.
I guess we could have them as hub module.
It can tightly be integrated later.
https://github.com/rwightman/pytorch-image-models

oke-aditya:
- We will use ross models, but not make package dependent on it. Nor we will provide fixes to his code / take any part of his code as our liability.

# Proposal 1: -

- Make us as hub to connect to his models

Dream API

```
model = hub.models.hrnet()
```

Models corresponds to rwightman's models. So we use his models, but not make it mandatory codebase. Also errors will resonate to his codebase. So we do not take that responsibilty.

oke-aditya:
- What we would take care is making them usable as backbones. That's the only code that we take responsibilty to maintain.
- These should provide a more flexible codebase.

lgvaz:
Related to #122 , should these models be inside `hub`? Or should we create another abstraction?

What are these models? Only backbones or also completely new algorithms?

Let's say we decide to integrate YOLO, how would that look like?

oke-aditya:
These are simply CNNs, provided by Ross, he also provides a few optimizers, layers too. We could take them too as a part.
Maybe we can rename them. They are definetely not bakcbones, we would need to add code to make them useful as backbones.

oke-aditya:
What Ross provides is set of CNNs, some layers, some optimizers (as of now). We need to add these to our hub. 

These CNNs -> Backbone conversion is code to be handled by us, as this would make it compatible with mantisshrimp.

User should access these models as backbones through our code.

lgvaz:
To fully understand how to properly do this, we need to specify how the integration of a custom "algorithm" would be like.

Sure, having new backbones and layers to experiment with torchvision `MaskRCNN` would be nice, but nothing changes much.

This is why I'm asking, how would we integrate YOLO? Or maybe, how to integrate efficientdet from Ross?

oke-aditya:
Its tricky, coz we want to take responsibility of only maintaing integration of custom algorithms. But not EfficientDet in whole. I am Unclear at this point. Need your thoughts.

oke-aditya:
We have to do something very similar to what we did to torchvision models.

lgvaz:
A good first step would be experimenting with these libraries first, to understand how they work.

[...]

PR #134, opened 2020-06-30
Adds Generic Backbone support

oke-aditya:
- This adds support to all 3 backbone cases.
- Refactored the Torchvision code too.
- We can actually just provide utils to load backbones as there are far too many and we would need to keep patching stuff.

I guess we can natively support torchvision out of the box as it is very slow to adopt new models.

Timm backbones can be used with these helper functions. People will not get FPN support for these networks as of now.


oke-aditya:
Example usage for the user.

```
resnet_net = timm.create_model("resnet50", pretrained=True)
ft_backbone = create_backbone_adaptive(resnet_net)
WheatModel(num_classes=2, backbone=ft_backbone)
```

If he tries some backbone that has no adaptive_avg_pool. This backbone creation works for any aribitary CNN too.

```
regnetx_008 = timm.create_model("regnetx_008", pretrained=True)
ft_backbone = create_backbone_generic(regnetx_008, out_channels=672 # This user should know)
WheatModel(num_classes=2, backbone=ft_backbone)
```

oke-aditya:
Closes #75 .

ai-fast-track:
I suggest we refactor `create_backbone_adaptive` by internally calling `create_backbone_generic` like here below:

```
def create_backbone_adaptive(model, out_channels: int = None):
    # Out channels is optional can pass it if user knows it
    # print(list(model.children())[-1].in_features)
    
    if out_channels is None:
        modules_total = list(model.children())
        out_channels = modules_total[-1].in_features
    
    return create_backbone_generic(model, out_channels=out_channels)
```

oke-aditya:
Yes. Nice refactoring.

oke-aditya:
@lgvaz @ai-fast-track any more suggestions. Tests are passing. 

We need a tutorial for this.

mantisshrimp/backbones/__init__.py
```diff
@@ -1,5 +1,3 @@
 from mantisshrimp.backbones.torchvision_backbones import *
-from torchvision.models.detection.backbone_utils import (
-    BackboneWithFPN,
-    resnet_fpn_backbone,
-)
+from torchvision.models.detection.backbone_utils import *
```
lgvaz, line 2: Don't import everything from torchvision, we don't control their name space, here we should only import what we need `BackboneWithFPN` and `resnet_fpn_backbone`

mantisshrimp/backbones/create_backbones.py
```diff
@@ -0,0 +1,33 @@
+__all__ = ["create_backbone_adaptive", "create_backbone_features"]
+
+from mantisshrimp.imports import *
+
+
+# Use this when adaptive avg_pooling is not available or you want to pass custom CNN
+# This can work with adaptive case too. So it is a bit generic case.
+def create_backbone_generic(model, out_channels: int):
```
lgvaz, line 8: I believe this method might lead to mistakes, it's not really generic, it just removes the last layer from a model. If a have a model that is not structure in the way of `[backbone, head]` it will fail silently.

For example, my model can me like so: `[conv1, conv2, conv3, linear1, linear2, linear3]`, cutting the last layer would only remove `linear3`

mantisshrimp/backbones/torchvision_backbones.py
```diff
@@ -2,80 +2,92 @@
 __all__ = ["create_torchvision_backbone"]
```
lgvaz, line 3: You should removed the commented lines

requirements.txt
```diff
@@ -5,4 +5,5 @@ pyyaml>=3.13
 fastcore>=0.1.17
 opencv-python>=4.2.0
 pytorch-lightning>=0.7.6
-torchvision==0.6.0
\ No newline at end of file
+torchvision==0.6.0
+timm
```
lgvaz, line 9: I'm still not sure this should be a requirement, let's wait a bit and see

oke-aditya:
Well we need to wait for [this](https://github.com/rwightman/pytorch-image-models/pull/175). For now it's just a method that worked quite well. Extracting features before AdaptivePooling layer. But I guess with this pull of Ross himself would integrate it and it would be a really clean way to do stuff.

oke-aditya:
Refer [this](https://colab.research.google.com/drive/1pk_kCgUMv-Hwm61Xmhz0LGkwn0j-QJkk#scrollTo=LPwRAYLP2yyy) colab notebook. It explains what is happening. Also we will wait for Ross to integrate the feauture extraction module.

mantisshrimp/backbones/create_backbones.py
oke-aditya: Yes, it will remove the last layer. But most CNN models are structured like this (for image classification) user might transfer his classification model into object detection model.

oke-aditya: Also refer the colab notebook. It explains what is happening.

mantisshrimp/backbones/__init__.py
oke-aditya: Changed.

mantisshrimp/backbones/torchvision_backbones.py
oke-aditya: Changed.

requirements.txt
oke-aditya: Removed.

ai-fast-track:
Great idea @oke-aditya for having created the [colab notebook](https://colab.research.google.com/drive/1pk_kCgUMv-Hwm61Xmhz0LGkwn0j-QJkk#scrollTo=0JEomEpa0WuR) to clarify things

oke-aditya:
Best part this stuff will work with retina net too later on. I hope 😅.

oke-aditya:
@lgvaz any suggestions ?

lgvaz:
> @lgvaz any suggestions ?

Yes, I'm starting the work with efficientdet, give me a little bit more time so I can give more helpful comments 😁

oke-aditya:
Not going to merge this. We have handled this 👍

PR #140, opened 2020-07-01
Fixes model tests

lgvaz:
Attempt of trying to run the model tests on github by reducing the image size


lgvaz:
@oke-aditya strangely, for custom backbones we're (sometimes) getting [this](https://github.com/lgvaz/mantisshrimp/pull/140/checks?check_run_id=827753019#step:6:228) error.

I tried searching but found no relevant information on how to fix it, any ideas?

The only changes I made were reducing the image size, so it can run on github action, they're still big enough for the tested models though.

oke-aditya:
I saw the error. It is torchvision interpolation issue. I'm unsure of why that should happen. The issue was open in [torch](https://github.com/pytorch/pytorch/issues/10654). I don't think it is fault of backbones as such coz it should work in any case.

PR #142, opened 2020-07-01
renames annotation parsed fields to plural

### LINKED ISSUE #128: Rename items to plural in parser
This is best described with an example, take a look at the parser tutorial [here](https://lgvaz.github.io/mantisshrimp/tutorials/parser.html)

Note that the method 'bbox' is actually returning a list of bboxes. A more appropriate method to the method would be `bboxes`.

This change also reflects other methods down the line, for example `MantisFasterRCNN.build_training_sample`

lgvaz:
closes #128


examples/detr_wheat.py
```diff
@@ -30,16 +30,16 @@ def height(self, o) -> int:
...
+    def bboxes(self, o) -> List[BBox]:
         return [self.bbox]
 
-    def area(self, o) -> List[float]:
-        return [self.bbox.area]
+    def areas(self, o) -> List[float]:
+        return [self.bbox.areas]
 
-    def iscrowd(self, o) -> List[bool]:
+    def iscrowds(self, o) -> List[bool]:
```
ai-fast-track, line 42: The plural of crowd is still crowd: Crowd is uncountable noun. All "crowds" should be reverted to crowd

lgvaz: Nice catch up!! Thanks!

This is a scary refactor, but I think it's necessary.

lgvaz: Before I start the renaming, is it a problem that `iscrowd` is going to be associated both with a single item and a list of items?

Even if `iscrowds` is not a correct spelling, maybe it's a good idea to leave like so for consistency?

ai-fast-track: It should be isCrowd. Crowd implicitly refers to many (equivalent to a list)

lgvaz: It seems `crowds` is not technically wrong [source1](https://ell.stackexchange.com/questions/24616/difference-between-a-crowd-and-crowds) [source2](https://ell.stackexchange.com/questions/153754/should-i-fill-with-crowd-or-crowds-in-this-context) [source3](https://simple.wiktionary.org/wiki/crowds).

Let's stick with that because it makes a lot of intuitive sense, e.g. you might want to do `for iscrowd in record['iscrowds']`

PR #144, opened 2020-07-02
removes internal transforms from rcnn models

### LINKED ISSUE #131: Rescale transforms and torchvision RCNNs
## 🐛 Bug
torchvisions `FasterRCNN` and `MaskRCNN` internally rescales the images via `GeneralizedRCNNTransform`.

This means that any resizing transform previously applied to the image (probably on the `Dataset` stage) will be unexpectedly  overriden.

It becomes really confusing if we should apply a resize transform via `Dataset` or via the model, ideally we want all transform to be applied at the `Dataset` stage. Going one step further, can this even be said about normalization?

## Solution 1
Changes the default of `min_size` and `max_size` in the model to 2 and 9999 respectively, practically making them non effective.

## Solution 2
Monkey patch `model.transform` to a function that returns the same items it receives (being careful with what are the expected output types)

---
Solution 1 seems like a better idea, it's less magical, we can still have `min/max_size` as parameters, we just change their defaults.

---
Should we do the same with normalize? Changing the mean to 0 and std to 1? It's them clear that if we want to normalize we should do it in the `Dataset` stage.

I'm thinking this because models that we integrate in the future might not have this internal transforms, and it will be very confusing to a new user why sometimes he has to explicity define normalization and sometimes not.

lgvaz:
closes #131

### DIFF
--- a/mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
+++ b/mantisshrimp/models/mantis_rcnn/mantis_faster_rcnn.py
@@ -4,7 +4,6 @@ from mantisshrimp.imports import *
 from mantisshrimp.core import *
 from mantisshrimp.models.mantis_rcnn.rcnn_param_groups import *
 from mantisshrimp.models.mantis_rcnn.mantis_rcnn import *
-from mantisshrimp.backbones import *
 
 
 class MantisFasterRCNN(MantisRCNN):
@@ -22,6 +21,7 @@ class MantisFasterRCNN(MantisRCNN):
         backbone: nn.Module = None,
         param_groups: List[nn.Module] = None,
         metrics=None,
+        remove_internal_transforms=True,
         **kwargs,
     ):
         super().__init__(metrics=metrics)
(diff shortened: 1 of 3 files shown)


oke-aditya:
Can I know what has been done as solution?

lgvaz:
Still a draft that I'm checking out 😅

But basically this:
```python
    def _remove_transforms_from_model(self, model: GeneralizedRCNN):
        def noop_normalize(image):
            return image

        def noop_resize(image, target):
            return image, target

        model.transform.normalize = noop_normalize
        model.transform.resize = noop_resize
```

oke-aditya:
- How does this work ? So we do not do any transforms and pass the image to the model ?
- If we remove Generalized RCNN transform then user cannot pass images which are of diffrent sizes I guess.
- Can you document an example and check its working with all cases ?

lgvaz:
All transforms should be done in the `Dataset` level, e.g. with albumentations

The new tutorial [here](https://lgvaz.github.io/mantisshrimp/tutorials/mantisshrimp_models/faster_rcnn_quickstart.html) show how its done

PR #153, opened 2020-07-06
Fixes requirements

oke-aditya:
- We don't need torchvision in requirements. Users should install it with PyTorch themselves as versions of PyTorch and torchvision go hand in hand

- We don't need NumPy. It will be included with torch.


oke-aditya:
- Due to the plotting utils update we need matplotlib compulosorily. Simply installing requirements won't run the model.

Also Pycoco tools is hardcore dependency so it causes trouble still.

Issue #137, opened 2020-06-30
Models refactor


lgvaz:
## Overview

As @ai-fast-track suggested, we could use a more functional approach for the methods currently present in `MantisModule` and more specifically in `MantisRCNN`.

The main source of confusion is that our models are currently doing two things: Defining the neural network and implementing the algorithm.

To be more specific, these are the methods our models currently require:
* forward: how to feed-forward the NN
* loss_func: given the preds and targets calculate the loss
* dataloader: given the dataset get the dataloader
* build_training_sample: convert from a prepared_record to something the model can understand. This works as an adapter between our data structures and the data structures expected by the NN.

The main point of discussion was the `.dataloader` method, claiming that it's not logical to have a method on the model that returns a `DataLoader`, and it's more logical to just have a function like "rcnn_dataloader" that return a `DataLoader` with the required modifications (like a custom collate_fn).

## Objective of the interface
Provide functions that ease out the integration of models/algorithms with other training frameworks. For example, any training framework will require the specification of a loss function, our library should be able to provide that.

---
**Note:**
Currently we're only working with torchvision, and this is biasing us of having everything into the model because this is the way they done it.

It's unnatural to separate those functions because the model is already tightly coupled in their implementation, while training, their model actually requires you to pass the targets to the `forward` method so then return you a list of losses instead of the predictions! (and this is very problematic, for instance, in validation we need to do one forward pass to get the losses and another to get the predictions)

**Suggestion:**
Similar to what we did when making an agnostic training api, I suggest that we first try to integrate another model/algorithm to the current structure before trying to come up with another API.

This would give us another point of view and make things clearer on how a more generic API should look like, forcing us into making better decisions.

Yolo could be a good addition to the library, compared to MaskRCNN, Yolo is an algorithm that trades a bit of precision for a lot of speed. We can also try to integrate Ross efficientdet (although at this moment I don't really understand how efficientdet works)

lgvaz:
This is going to be one of the most important interfaces of mantisshrimp so it has to be very clear and intuitive!

I encourage everyone to give their suggestions so we can settle on an interface that is modular and straight forward! 🔥

lgvaz:
@ai-fast-track , @oke-aditya , @philtrade

lgvaz:
Currently, we have a single function for getting a dataloader, called on `model.dataloader`. It's the case for `RCNN` that the training and validation dataloaders are identical, **but** for efficient-det they are a slightly different...

[...]

PR #155, opened 2020-07-06
implements efficient det

### LINKED ISSUE #149: Add efficientdet support
## 🚀 Feature
Let's start working on adding support for efficientdet, specifically, let's work on [this](https://github.com/rwightman/efficientdet-pytorch) implementation.

A good starting point is [this](https://www.kaggle.com/shonenkov/training-efficientdet) kaggle kernel, which is using efficient det to finetune in the wheat dataset.

---
Related to #137

lgvaz:
While we wait for ross answers and advice on how to integrate efficient-det, I already started implementing some very basic tests. This is called test-driven development, you first write what you expect and only then you implement the code to achieve that.

I'm using [this](https://www.kaggle.com/shonenkov/training-efficientdet) kaggle notebook as a guideline. I used that notebook to understand what we need to output at the dataset stage and on the dataloader stage, and then I wrote code to replicate that.

@oke-aditya Do you agree with the expected outputs? There is still that confusion about the BBox format (`xyxy` or `yxyx`), the notebook uses `xyxy` so I went with that. I asked that question to Ross so we will know the answer very shortly.

---
closes #149


lgvaz:
I originally said I would first integrate efficient and only then work on #137, but I'll start on that already while we wait for Ross to respond.

lgvaz:
First working version but very very very dirty

lgvaz:
@ai-fast-track @oke-aditya 

I have some notebooks (modified from the quickstart) working to show the new api:
* [faster_rcnn](https://colab.research.google.com/drive/1meOFxlRsXIyakYsabc8R7m_LDel4fveU?usp=sharing)
* [efficientdet](https://colab.research.google.com/drive/1e6Rb10OFTq-3zuP6h4AeJSiK4yD7S4Cw?usp=sharing)

I still don't know if the efficientdet implementation is working correctly, I'm training a model for that

Keep in mind we cannot directly compare faster_rcnn and efficiendet because they're using slighty different transforms

lgvaz:
This PR is still not ready for merging, I still need to work on `efficient_det.predict`

lgvaz:
@oke-aditya @ai-fast-track 

The model is training successfully, the loss is going down, but I'm unable to get reasonable predictions

Can you take a look at the notebook [here](https://colab.research.google.com/drive/1e6Rb10OFTq-3zuP6h4AeJSiK4yD7S4Cw?usp=sharing)?

Code is still very dirty, but at the prediction section there is the code I'm trying to hack around to make prediction work...

Results seem to always stay the same no matter how much I train the network, I think there must be a very obvious bug that I'm failing to see at this point

oke-aditya:
K. I'm Having a Look. Let me try to train it once.

oke-aditya:
I'm unable to run the above notebook, can you have a look at errors.
1. Is the Lightning adapter correct (RCNN Lightning adapter for efficientdet ?)
2. Reshaping the images to 512, 512 is needed for the model but it destroys the aspect ratio of images, it can cause trouble in predictions as anchor boxes learnt might differ.
3. Learning rate should be done with Adam, SGD is not converging, it is unstable training.
4. While predicting are we resizing the Bounding boxes too properly? This might be a source error since we predict on images of size 512, 512, then resize them. Are we resizing the bounding boxes too properly?
5. I'm trying with D7. If it works well then maybe we need this model. While debugging we need to start from bigger models as once any one of them works all will have same pipeline.

lgvaz:
1. I did not tested lightning still (the code is unchanged from the older tutorial), can you try with fastai?
2. Agree! But we still should be able to see resoanable results right? I mean, this can't be the source of the problem
3. Addressed by 1
4. Can you elaborate? I'm using the images from valid_ds to predict, which are already scaled to 512, so we don't have to worry about this right now
5. Cool! let me know if you get better results

lgvaz:
This PR contains a lot of other changes that are not related to efficient det (an entire model refactor for example). So I'll merge this and open an issue for fixing efficientdet

oke-aditya:
1. Fastai too didn't work work for me. I don't know why. 
4. We are resizing and predicting images, that is fine. My doubt is on show_pred we are predicting on true images which are not scaled, so maybe there is resizing issue. We need to predict on the valid_ds image, draw bbox on the image (so that it too gets resized) then resize if needed. I am not sure how matplotlib is handling that. If it is on canvas which does not get resized then that is an issue,

Rest all LGTM, we can surely go ahead for now and fix issues later.

lgvaz:
1. Can you share the error?
2. In the show_preds we're also using the scaled images, from here:
```python
samples = random.choices(valid_ds, k=2)
images = [sample["img"] for sample in samples]
```
Those come from valid_ds, and those are used in `show_preds` as well

oke-aditya:
Seems to be working. I guess I messed up Fastai becoz I tried lightning first. It is slow to train. Currently I am working in colab [here](https://colab.research.google.com/drive/1Aoxzsdku_zocjtM211FIdq0qdvGXPKbL?usp=sharing)

PR #162, opened 2020-07-09
Fix setup

oke-aditya:
- Have a close look. Let me know if changes are needed.
- We will merge this when we reach 
```
pip install mantisshrimp[minimal]  should give you minimal installation without any engine
pip install mantisshrimp[fastai]  should give you mantis + fastai
pip install mantisshrimp[pl]  should give you mantis + lightning
pip install mantisshrimp[all]  should install entire package. All engines etc.
``` 

this strategy.


oke-aditya:
Well was this done correctly? I mean is the above way of installing going to work after this PR?

Issue #135, opened 2020-06-30
Mantisshrimp Milestones


ai-fast-track:
# 🚀 Milestones

This lists the pre-requisites for our **Milestones**:

## Milestone 1
**Parsers**: Implementing the following parsers
	- [x] COCO Parser
	- [x] VOC Parser

**DL Engines**: Bridging Mantisshrimp to different DL libraries
	- [x] Fastai Integration
	- [x] Pytorch-Lightning Integration

**Models**: 
	- [x] FasterRCNN
	- [x] MantisMaskRCNN

**Backbones**: Connecting different backbones to Mantisshrimp
	- [x] Backbone Generic 
	- [x] Backbone Adaptive 
	- [x] Backbone Features 
	- [x] torchvision Connection

**Models Hub**: Loose coupling with different models implementation through referencing to original models repos
	- [x] Detr

**Parsers Hub**: Community Parsers
	- [x] Wheat Parser
	- [x] PennFundanParser

**Datasets**: List of datasets used in tutorials
	- [x] Pets
	- [x] Wheat
	- [ ] Fridge objects

**Tutorials**: 
	- [ ] Getting started: FasterRCNN + Pets dataset + Fastai + Pytorch-Lightning + Visualization (showing images, boxes, and labels
	- [ ] Advanced Tutorial: similar to the previous but using Wheat dataset + uses different backbones
	- [ ] How to train your own dataset: Step by step tutorial - Uses Fridge objects dataset

**Documentation**: 
	- [ ] Documentation generated using  `.py` modules
	- [ ] Documentation generated using the different tutorials notebooks
	- [ ] Creating some visual illustrations

## Milestone 2
It is Milestone 1 + the following pre-requisites:

**Deployment**: 
	- [ ] Using Streamlit: Locally + in the cloud

**Datsets**: Add other datasets such the following:
	- [ ] Blood cells dataset (BCCD)
	- [ ] Plant disease
	
**Logging**: Integrating some logging tools
[ ] Wandb

lgvaz:
Can you create this as a project? (or as a milestone even, I just find projects more organized because of the cards)

lgvaz:
Moved to the #Milestone1 project

Issue #83, opened 2020-06-16
Interactive visualization of data


lgvaz:
## 🚀 Feature

We need an interactive way to visualize data, we can interactively visualize bounding boxes and masks with wandb like described [here](https://www.wandb.com/articles/bounding-boxes).

We can also use streamlit, I'm not sure which is cooler.

Wandb draws more my attention because we can also use it for visualizing training statics (in real time), something that would be very very helpful, is to show the predictions of the model after every epoch.

This would provide strong insight on what the model is correctly learning and what it's fail to learn,  I believe a tool like this quickly becomes invaluable for anyone training models (it actually surprises me that this simple feature is not widespread at this point)

lgvaz:
@paras-jain is this something you would like to work on?

paras-jain:
Yes sure! streamlit makes deployment very easy and wandb helps in tracking and optimization, we should not miss out on either. I will look into it and then work on #84.
Has any [previous](https://github.com/lgvaz/mantisshrimp/issues/39#issuecomment-636639684) work been done?

lgvaz:
Not really, we talked about it multiple times, but never found the time to implement it 😅

oke-aditya:
> 
> 
> Yes sure! streamlit makes deployment very easy and wandb helps in tracking and optimization, we should not miss out on either. I will look into it and then work on #84.
> Has any [previous](https://github.com/lgvaz/mantisshrimp/issues/39#issuecomment-636639684) work been done?

Something to get strated
https://github.com/streamlit/demo-self-driving

FraPochetti:
Taken care of by @fstroth 's dashboard

Issue #39, opened 2020-05-30
Tutorials and Examples


lgvaz:
## 📒 Tutorials
Tutorials are in `.ipynb` format, explaining each step of the process, really detailed, not production like.

**Core**
- [x] [parser](https://lgvaz.github.io/mantisshrimp/tutorials/parser.html)

**Object detection**
- [x] [wheat](https://lgvaz.github.io/mantisshrimp/tutorials/wheat.html)
- [x] [wheat with detr](https://lgvaz.github.io/mantisshrimp/tutorials/hub_detr_finetune_wheat.html)

**Segmentation**
- [ ] [penfundan](https://pytorch.org/tutorials/intermediate/torchvision_tutorial.html)

**Keypoints**

## 📓 Examples
Examples are be in the `.py` format, more production oriented. Ready to be run with arguments from the command line and easy to integrate with wandb sweeps and alike.

**Object detection**
- [ ] [wheat](https://www.kaggle.com/c/global-wheat-detection)

**Segmentation**

**Keypoints**

---
Is there a new tutorial or example you would like to add? Comment below and we talk about it 😁

Once we agree, create an **Tutorial or Example request** issue (use the template) and I'll edit this post with your new cool example!

lgvaz:
It would be a cool addition to show how to do a wandb sweep in the whet example

chho-work:
Certainly it would be, Wandb supports Pytorch Lightning: https://docs.wandb.com/library/frameworks/pytorch/lightning

I will work on few examples with different datasets!

lgvaz:
It might be a good idea to separate this into `Examples` and `Tutorials`.

Tutorials would be in `.ipynb` format, explaining each step of the process, really detailed, not production like.

Examples would be in the `.py` format, more production oriented. Ready to be run with arguments from the command line and easy to integrate with wandb sweeps and alike.

What do you think @chho-work ?

chho-work:
Yes, it would be great to have both available, examples and tutorials, each with their corresponding formats.

chho-work:
I found this pretty interesting!  I think we can use them in tutorials and examples.
https://github.com/mseg-dataset/mseg-api

tazu786:
Hi, sorry for my probably silly question, but when I run 
>run_detr(args, train_dataset, valid_dataset)

in the tutorial for fine tuning the detr, I get
> TypeError: main() takes 1 positional argument but 3 were given

How can I pass the datasets to the main(args) of detr?

Thank you

[...]

PR #168, opened 2020-07-10
Add streamlit readme

oke-aditya:
Adds streamlit to readme.
Please check if the link works.


oke-aditya:
Helps in #135 #39 #83

README.rst
```diff
@@ -258,6 +258,21 @@ Quick Example: How to train the **PETS Dataset**
    trainer = Trainer(max_epochs=3, gpus=1)
    trainer.fit(light_model, train_dl, valid_dl)
 
+Streamlit Demo
+--------------
+
```
ai-fast-track, line 263: It would be good to add How to install streamlit, something like this:
--------------
If streamlit is not already install, run the following command, from the terminal to install it:

`pip install streamlit`

oke-aditya: Added.

lgvaz:
Very nicee!!!

@oke-aditya can you clean the streamlit code before we add this to the readme? 

Currently there are a lot of commented code lines, and some general refactor could be done as well, for instance, you have very helpful comments explaining what is the purpose of the function, those should be docstrings.

Keep in mind that this is going to be our front page for deployment, so the code needs to be as clean as possible

oke-aditya:
Yes I will clean it up. Some final touch up is needed.

oke-aditya:
- Done ✔️ You can have a look [here](https://github.com/oke-aditya/mantisshrimp_streamlit)
Maybe a bit more refactor can be done, But for now I guess this should be good enough.
@ai-fast-track can you have a look at it ?
Let me know if something needs more change.

lgvaz:
Alright, much better! But there are still some things

* Repeating the big list OBJECTS_TO_DETECT  in both config.py and app.py
* config.py: `# IMG_SIZE = (224, 224) # If you need a image size` what does this mean?
* duplicate code between `app.py` and `model.py`
* duplicate code between app.py and object_detection_app.py
* duplicate code in utils.py and other files

lgvaz:
Also, are you doing imagenet normalization? You can use `normalize_imagenet`

oke-aditya:
Let me clear the code duplication issue.
app.py is a standalone app that should run when user does `streamlit run https://raw.githubusercontent.com/oke-aditya/mantisshrimp_streamlit/master/app.py`

This file cannot have relative imports etc. So I need to paste all the code in src into here. 
The src folder is provided so that user can design his own app easily, modularize it into files and easily hack the script to make a new app.

So ignore the duplication with app and others.
IMAGE_SIZE I will correct it.
ImageNet normalization, I need it but where should I add it let me know. It works without it though.

ai-fast-track:
@oke-aditya, I think we need to have just one single `app.py`. I understand we had the `src/...app.py` before the prospect of having the remote `app.py`, and now we can eliminate code duplication.

I just tested the 2 options, they both work: So one code base, let's keep it DRY 😃  .

**Option 1: Local**
`streamlit run app.py`

**Option 2: Local**
`streamlit run https://raw.githubusercontent.com/oke-aditya/mantisshrimp_streamlit/master/app.py`

Some changes:

I think this should be changed in `app.py`
`APP_NAME = "app.py"` instead of  `"src/object_detection_app.py"`

In line 217, `mantiss` should be replaced by `mantis` (one s)

In line 258: add **?** to `st.sidebar.title("What to do")`

in line 307:  Replace **You** by **Your** `caption=f"You amazing image has shape {image.shape[0:2]}",`

In `Info.md` file:

Replace To run click on Run app in sidebar. **by** To run the app, select **Run app** from the combo box on the sidebar (on the left).

oke-aditya:
ok. I will make the suggestd changes. If I keep only the app.py. Should I keep files such as src/utils.py, src/model.py etc ?

ai-fast-track:
I think It would be better to remove them, and in the `app.py` highlight the different sections corresponding to each original files (model, utils, etc.) in order to visually distinguish each of those original components.

Something as simple as this (or similar):
################################################################################################
#
# Model Section
#
################################################################################################
#
# Use the mantisshrimp Pytorch model here.
# Simply create the model and get its predictions.

Great Job!

oke-aditya:
Ok. I will remove that too from master branch. Instead have master branch only with app.py from where users can run.
And have this modularized code in the dev branch. This enables me to easily fix up the code when required, and add features if needed.

oke-aditya:
Done. I have updated the master branch. It has only app.py. I tested it. It is working. Also, we on dev branch the modularized code which a user might need.

PR #171, opened 2020-07-12
removed duplicated getting-started notebook from the root path

ai-fast-track:
Clean up: I removed duplicated 01_getting-started notebook from the root path.


ai-fast-track:
I just deleted the duplicated getting started notebook  from the root folder

PR #174, opened 2020-07-12
Add support for cvbp-fridge dataset

paras-jain:
I did not find multiple objects of same class hence removed `num_objs`. Please review.


ai-fast-track:
Great! I'm looking into it.

PR #178, opened 2020-07-15
added transforms for PETS dataset in pets folder

ai-fast-track:
I added a `transforms.py` file in the `pets `folder that defines the following method:

`aug_tfms_albumentations_pets(train=True)`

It will be called like this (in 01-getting_started notebook for example)

```
from mantisshrimp.datasets.pets.transforms import *
# Use predefined transforms or create your own
train_tfms = aug_tfms_albumentations_pets(train=True)        
valid_tfms = aug_tfms_albumentations_pets(train=False)
```

### DIFF
--- a/mantisshrimp/datasets/pets/transforms.py
+++ b/mantisshrimp/datasets/pets/transforms.py
@@ -4,36 +4,29 @@ __all__ = [
 
 from mantisshrimp.imports import *
 from mantisshrimp.core import *
-from mantisshrimp.transforms.albu_transform import *
-from mantisshrimp.utils.utils import *
-
-import albumentations as A
 
 
-
-def aug_tfms_albumentations_pets(train=True):
-    """ Composes a Pipeline of Albumentations Transforms for PETS dataset
-
-    Args:
-        train: flag to select transforms specific to either train or validation stage
+def train_albumentations_tfms_pets ():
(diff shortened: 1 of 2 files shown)


mantisshrimp/datasets/pets/transforms.py
```diff
@@ -0,0 +1,59 @@
+__all__ = [
+    "aug_tfms_albumentations_pets"
+]
+
+from mantisshrimp.imports import *
+from mantisshrimp.core import *
+from mantisshrimp.transforms.albu_transform import *
+from mantisshrimp.utils.utils import *
+
+import albumentations as A
```
lgvaz, line 10: Can you move this to inside the function? Or else it will be a requirement to have albumentations to use pets

mantisshrimp/datasets/pets/transforms.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+
+def aug_tfms_albumentations_pets(train=True):
```
lgvaz, line 14: I think we can remove `aug`, since if `train=False` no augmentation transforms are used.

What do you think of the idea of having two functions? `train_albumentations_tfms_pets` and `valid_albumentations_tfms_pets`? That also eliminates the need of `if` inside the function.

PR #180, opened 2020-07-16
implements img_pad_stack transform

### LINKED ISSUE #177: Batch images of different sizes
## 🚀 Feature
With the landing of #155 batch transforms will be a possibility.

We can write a transform that pads and batches (into a single tensor) images of different sizes. This is what torchvisions RCNNs does internally (this is why they support a list of images of different size as input).

We should implement everything in numpy, for inspiration, here is how torchvision does it:

```python
    def batch_images(self, images, size_divisible=32):
        # type: (List[Tensor], int)
        if torchvision._is_tracing():
            # batch_images() does not export well to ONNX
            # call _onnx_batch_images() instead
            return self._onnx_batch_images(images, size_divisible)

        max_size = self.max_by_axis([list(img.shape) for img in images])
        stride = float(size_divisible)
        max_size = list(max_size)
        max_size[1] = int(math.ceil(float(max_size[1]) / stride) * stride)
        max_size[2] = int(math.ceil(float(max_size[2]) / stride) * stride)

        batch_shape = [len(images)] + max_size
        batched_imgs = images[0].new_full(batch_shape, 0)
        for img, pad_img in zip(images, batched_imgs):
            pad_img[: img.shape[0], : img.shape[1], : img.shape[2]].copy_(img)

        return batched_imgs
```

It's really not hard, just find the max dimensions and pad right and bottom with zeros. (Ross also does this in efficientdet, and he said that padding with imagenet mean gives better results).

lgvaz:
closes #177 , the result looks like this:

image

### DIFF
--- a/mantisshrimp/transforms/__init__.py
+++ b/mantisshrimp/transforms/__init__.py
@@ -1,2 +1,4 @@
 from mantisshrimp.transforms.transform import *
 from mantisshrimp.transforms.albu_transform import *
+
+import mantisshrimp.transforms.batch_tfms
--- a/mantisshrimp/transforms/batch_tfms/__init__.py
+++ b/mantisshrimp/transforms/batch_tfms/__init__.py
@@ -1,0 +1,2 @@
+from mantisshrimp.transforms.batch_tfms.batch_transform import *
+from mantisshrimp.transforms.batch_tfms.img_pad_stack import *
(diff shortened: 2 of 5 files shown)


lgvaz:
Is also possible to pass a custom value to pad with, e.g. imagenet mean:

image

Issue #209, opened 2020-07-24
Release 0.1.0


oke-aditya:
## 🚀 Feature

This might seem a bit early. But we are now almost close to First Release 0.1.0.
We need to follow a release method that will enable us to create a release cycle and properly keep our job done.
This is the method PyTorch follows and is very efficient.
1. Cut a branch for release a week prior, tag it as release/version number this case release/0.1.0. Update the setup.py and setting.ini etc files. This will enable releasing on the new version.
2. Create an issue for release. Tag it as [Release 0.1.0] DO NOT MERGE.
3. Now PRs do not get merged to release branch without approval or cherry-picking. Any further PR to this release must be linked to this Release issue, only then we can think of merging. Only essential and important PRs can be merged. We can Pinup this issue so people know about it.
4. This ensures that the master remains stable and we can keep developing on the master without worrying about the release branch it will stay independent. All CI checks need to be passing on the release branch. No CI failure can be accepted.
5. Finally while releasing, we need to create a PR on release branch with a tag. We need to add a GitHub action for releasing the package.
6. Merge the same the release branch into docker Deploy. It will also deploy the docker image for our repo.
7. Do not delete the release branch. Keep that branch. It can stay for long.

To do before this release once this stuff is done we can close this issue. 

- [ ] Fix the docker deploy action and check it once. Issue #127 
- [x] We need to create proper requirements and configure the settings.ini and setup.py properly. Issue #182.
- [x] Fix up nasty imports which create lot of problems. Issue #181 #184 .
- [ ] Create CI files for Python Package releasing.

Once these are fixed and milestones are reached. We can cut a release branch. And then follow the above steps. 
Note that we need to keep a weeks gap at least when we cut the release branch and actually release it so that we can make last-minute very important fixes.

Issue #182, opened 2020-07-16
Create Seperate PyPI Packages


oke-aditya:
## 🚀 Feature

While deploying we need to reduce our size, requirements and also the necessary dependencies,
Also since engines are independent we can reduce the footprint of our code here as well.

This is what we decided.

```
pip install mantisshrimp # This is not bare for ease of new users. It downloads all
pip install mantisshrimp[bare]
pip install mantisshrimp[lightning]
pip install mantisshrimp[fatai]
```
This will require to edit `setup.py` and the `requirements.txt` and `settings.ini`

The key step should be #181. Helps #209.

oke-aditya:
Some issues that are left over from #217 

1. Separate CI for engines. This is CI for All engines and complete installation (some extra unnecessary stuff too). And a seperate CI for no engine deployment ready mantisshrimp. We need to mark our tests in order to support these.
2. How to add Ross Efficient det to setup.py. It is not getting bundled. I have patched it in CI which is not correct.
3. CI action to deploy this package with tag push.
4. Fix up the imports !!! they will help in Pruning Packages. Maybe some we do not need in minimal installation.

oke-aditya:
Guess we are done with this. I am closing it.

Issue #181, opened 2020-07-16
Stop importing everything everywhere 


oke-aditya:
## 🚀 Feature

Ease of Programming vs Deployment is here.
Using from `mantisshrimp.imports import *` this causes deployment issue. It would lead to the import of unnecessary libraries and immediately give errors.

E.g. when the user needs only inference, he starts importing the model. We have `mantisshrimp.imports import *` which will try importing fastai and lightning. Which will throw an error.

Same thing for utils `mantisshrimp.utils import *`.

This is a painful refactor but this causes the biggest issue while deploying. It would be hard to program with this but if we can limit the imports it will be easy and consume less memory footprint too. We cannot create PyPI packages.
```
pip install mantisshrimp # This is not bare for ease of new users. It downloads all
pip install mantisshrimp[bare]
pip install mantisshrimp[lightning]
pip install mantisshrimp[fatai]
```

lgvaz:
This is not happening anymore, soft dependencies are already in play

Closing this =)

PR #217, opened 2020-07-24
Repackages code for pypi

oke-aditya:
Updates for better PyPi packaging. Helps #182. #209

So now how the installation goes.
Note that

If we are at the repo root. then we do `pip install .` otherwise on PyPi we would just replace `.` with `mantisshrimp`

With this update, we will be able to install engines separately through `setup.py` itself. This follows the HuggingFace strategy exactly. Which we aimed for.
 
- To install mantisshrimp raw. without any engine
`pip install mantisshrimp` or if we have cloned and we are installing `pip install .`

- To install with Fastai.
`pip install mantisshrimp[fastai]` or if we have cloned and we are installing `pip install .[fastai]`

- To install with Pytorch-lightning
`pip install mantisshrimp[pytorch-lightning]` or if we have cloned and we are installing `pip install .[pytorch-lightning]`

To install completely with both the engines
`pip install mantisshrimp[all]` or if we have cloned and we are installing `pip install .[all]`

Note that we cannot provide the all option as default as these packages are built incrementally and that's what even HF follows.

We expect the user to have torch and torchvision pre-installed otherwise we do that for them.

I see our CI is kinda broke. It installs from `requirements.txt` and `requirements-extra.txt`. This will not work henceforth. If we are packaging this code then we should have `pip install .[all]` in our repo base and check for the tests. `requirements.txt` and `requirements-extra.txt` can completely be deleted then.

This CI change will ensure that what we build in `setup.py` is what we test and will be what we will deploy. Also it will ensure that correct requirements are listed in `settings.ini` and there is no discrepancy.

I will raise PR for CI change. Then re-run this to check for it.

This current installation method. I have checked locally all seem to be working good. Also, I will add a CI to deploy to PyPi when we tag a release later.

### DIFF
--- a/.github/workflows/ci-all-testing.yml
+++ b/.github/workflows/ci-all-testing.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: Tests
+name: tests
 
 on:
   push:
--- a/.github/workflows/ci-all-testing.yml
+++ b/.github/workflows/ci-all-testing.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: CI All Test
+name: Tests
 
 on:
   push:
--- a/.github/workflows/ci-all-testing.yml
+++ b/.github/workflows/ci-all-testing.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: CI Test
+name: CI All Test
 
 on:
   push:
@@ -28,7 +28,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install flake8 pytest pytest-cov
         pip install torch==1.5.1+cpu torchvision==0.6.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
-        pip install -e .
+        pip install -e .[all]
         pip install git+git://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI
     - name: Lint with flake8
       run: |
--- a/.github/workflows/ci-all-testing.yml
+++ b/.github/workflows/ci-all-testing.yml
@@ -30,6 +30,7 @@ jobs:
         pip install torch==1.5.1+cpu torchvision==0.6.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
         pip install -e .[all]
         pip install git+git://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI
+        pip install git+git://github.com/rwightman/efficientdet-pytorch.git
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
--- a/settings.ini
+++ b/settings.ini
@@ -25,7 +25,9 @@ license = apache2
 status = 2
 
 # Optional. Same format as setuptools requirements
-requirements = fastcore torch torchvision
+requirements = fastcore tqdm matplotlib torch torchvision
+pytorch_lightning = pytorch-lightning
+fastai = fastai2
 # Optional. Same format as setuptools console_scripts
 # console_scripts =
 # Optional. Same format as setuptools dependency-links
--- a/settings.ini
+++ b/settings.ini
@@ -25,7 +25,7 @@ license = apache2
 status = 2
 
 # Optional. Same format as setuptools requirements
-requirements = fastcore tqdm matplotlib torch torchvision
+requirements = fastcore tqdm matplotlib cython opencv-python omegaconf ipykernel git+git://github.com/rwightman/efficientdet-pytorch.git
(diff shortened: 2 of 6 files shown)


settings.ini
```diff
@@ -25,7 +25,9 @@ license = apache2
 status = 2
 
 # Optional. Same format as setuptools requirements
-requirements = fastcore torch torchvision
+requirements = fastcore tqdm matplotlib torch torchvision
```
lgvaz, line 28: Can we take out torch and torchvision from here and specify it should be previously installed?

lgvaz:
> Note that we can provide the all option as default

That seems better, can we go with that? 

And then have a [minimal] tag

lgvaz:
There are some missing packages to install, for example, fastai requires `ipykernel`

We also need opencv

Also, how can we handle other additional deps like efficientdet?

---
A request, before we merge this, can you start from an empty env and confirm that we can run `getting-started` with faster_rcnn and efficientdet with both training engines?

You should actually create one env for each engine and confirm they are working individually

lgvaz:
All should also contain additional packages like `albumentations`

settings.ini
oke-aditya: We can take that out. But we need to ensure that the user has these two installed. If they aren't then how would we handle?

oke-aditya:
> 
> 
> > Note that we can provide the all option as default
> 
> That seems better, can we go with that?
> 
> And then have a [minimal] tag

I did a typo. Note that we **cannot** provide the all option as default. So by default the installation is Minimal.

oke-aditya:
> 
> 
> There are some missing packages to install, for example, fastai requires `ipykernel`
> 
> We also need opencv
> 
> Also, how can we handle other additional deps like efficientdet?
> 
> A request, before we merge this, can you start from an empty env and confirm that we can run `getting-started` with faster_rcnn and efficientdet with both training engines?
> 
> You should actually create one env for each engine and confirm they are working individually

All these sub-requirements like `ipykernel` would be installed by fastai2 itself. For efficientdet and other repos I think the default installation should come with all of these. Even when deploying you need to instantiate the models.

settings.ini
oke-aditya: While testing Torch triggered a wrong package, pytorch should be done by the user. We will automate fastai and lightning. Agreed and changed. Torch and torhvision go hand in hand so better user handle from PyTorch.

oke-aditya:
I am testing stuff locally. Also I will edit the current CI so that stuff works perfectly.

oke-aditya:
> 
> 
> All should also contain additional packages like `albumentations`, `pytest`

Is albumentations mandatory for user ? He can use out of the box torchvision transforms too right?

oke-aditya:
We need a fix for #181 Without this it is becoming very hard to limit the dependencies. Everything is imported everywhere so packaging might not work well. Also I need to write seperate tests for users using only lightning and users using only fastai engine.
We need to refactor the tests folder into these two subfolders. Then we can do proper Testing.


As of now I am updating the settings.ini with suggestions. But we need the above stuff.

oke-aditya:
As of now I have passed the tests with this update of dependencies. I will add a PR for CI

oke-aditya:
Yeah, so this builds with the new CI test itself. So I guess with this we will have both stuff done.

oke-aditya:
We need revamp of imports in tests folder as well. It is again importing everything unnecessary.
I will make sure that this packaging works well for now. This will help us while we revamp the unnecessary imports. 
In a later PR I will try to Prune the dependencies. For now this seems easy and feasible solution. I do not want to add many changes and keep breaking CI or requirements.

oke-aditya:
CI Test fails because of Ross efficient Det. I am unable add this to `setup.py` as it is a git repo. `setup.py` expects a egg file or pypi package. For now I am by passing this by adding it to CI testing file itself as bash command.

We need to fix this, if some-one can let me know how to package such github repos in `settings.ini` or in `setup.py` it would be easy fix.

Otherwise, we need help from Ross or build an egg file from his repo and ship it.

oke-aditya:
All CI pass. I have patched and fixed stuff.

Issues that still pertain: -
1. Separate CI for engines. This is CI for All engines and complete installation (some extra unnecessary stuff too). And a seperate CI for no engine deployment ready mantisshrimp. We need to mark our tests in order to support these.
2. How to add Ross Efficient det to setup.py. It is not getting bundled. I have patched it in CI which is not correct.
3. CI action to deploy this package with tag push.
4. Fix up the imports !!! they will help in Pruning Packages. Maybe some we do not need in minimal installation.

oke-aditya:
Thoughts @paras-jain @ai-fast-track @lgvaz

.github/workflows/ci-all-testing.yml
```diff
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: tests
+name: CI All Test
```
lgvaz, line 4: just call it tests for now

lgvaz:
Btw, do we currently have the default `pip install mantisshrimp` that installs everything?

oke-aditya:
No, but we have option `pip install mantisshrimp[all]` that installs everything. We cannot make `pip install mantisshrimp` to install all

.github/workflows/ci-all-testing.yml
oke-aditya: K renamed 👍  just keeping stuff as `Tests` Title looks easier on the eyes.

PR #221, opened 2020-07-25
Fixing the CI for packaging

oke-aditya:
Removes the need of requirements.txt and requirments-extra.txt. Our package should work from pip install -e . itself.


oke-aditya:
Maybe I can add this CI tests to the older PR #217 . I am closing this for that.

PR #223, opened 2020-07-26
Revamp CI for packaging

oke-aditya:
- With the current change I added CI for more OS systems. It is good idea to build on multiple OS before publishing.
- This technique is taken from PyTorch Lightning.

- I have added CI for releasing to both PyPi test server and PyPi official. This is again taken from Lightning

### DIFF
--- a/.github/workflows/ci-all-testing.yml
+++ b/.github/workflows/ci-all-testing.yml
@@ -12,9 +12,11 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


oke-aditya:
Why aren't all CI checks happening? This will result in a broke CI.

oke-aditya:
Retrying this by pushing into a new PR. This is utter bad GitHub. They should atleast throw an error if the CI yml has issue. But it must get triggered. Poor GitHub.

PR #225, opened 2020-07-26
Add publish ci

oke-aditya:


oke-aditya:
Bad Github. I am creating a new branch and adding this stuff. Github actions won't allow edits so easily.

PR #226, opened 2020-07-26
Adds new CI, Covers Multiple OS, Python Versions and Packaging

oke-aditya:
- With the current change I added CI for more OS systems. It is good idea to build on multiple OS before publishing.

- This technique is taken from PyTorch Lightning.

- I have added CI for releasing to both PyPi test server and PyPi official. This is again taken from Lightning

- We need a test to build package as well. And test PyPi releasing with a dummy release.


oke-aditya:
``` E   AttributeError: module 'torch.distributed' has no attribute 'group'```

I was too able to reproduce this error on my windows Machine
It arises from conftest.py code in our repo
Rest all, Linux, MacOs build for python 3.6, 3.7 and 3.8
This is a problem if `torch.distributed` is not initialized properly. Which is strange coz it builds on linux and not on windows.

**Comment by oke-aditya** [comment]
@lgvaz can you fix up the windows build issue of efficient det repo ? Once that is solved and I fix up the packaging We can then add these CI changes.

Helps in #182 #99 #135

**Comment by lgvaz** [comment]
@oke-aditya This is very very hard for me to solve, because, well.. I don't have windows hahahhah

Does windows support multi-gpu after all?

**Comment by oke-aditya** [comment]
Windows does support `torch.distributed` the issue maybe I am using cpu only which might not ship it.

**Comment by oke-aditya** [comment]
I'm facing a very small issue with Packaging install CI.
What it does is checks if the package can be installed properly from the wheel file.

Current scenario: -
I built the wheel, it gets saved in the dist folder. I cannot install that wheel using `pip install dist/`
More issues are it is version specific wheel name is `mantisshrimp.0.0.8.whl`

Possible solutions: -
Somehow if I can name the wheel as `mantisshrimp-test.whl` then I can install it by `pip install dist/mantisshrimp-test.whl`
Or some way to install wheel like *.whl. I am unable to do that as of now.

Other possible solution: -
Run `pip install .` Install the setup. This isn't a good thing, if we want to test the`.whl`.

**Comment by oke-aditya** [comment]
So Windows CI error is linked to https://github.com/pytorch/pytorch/issues/42095.
I guess torch itself does not support distributed GPU / CPU on windows but maybe soon it will. Let's see if there is a workaround or for now we can drop support for windows altogether (sad me).

**Comment by oke-aditya** [comment]
AAh the install bug was fixed with `MANIFEST.in` file. Why did we exclude it anyways ? It was there in nbdev template. We should have kept it.

**Comment by oke-aditya** [comment]
Again the build failed coz we tried to `import pandas as pd`. We are getting hit by #181 again and again.
I am removing pandas from init file. I guess we do not need it.

**Comment by oke-aditya** [comment]
Nice, we are getting there. New issue. We import PyTorch Lightning. Note we do not package this with our repo. It is not in our requirements as well, it is in additional requirements. So when we build the wheel, we do not package this dependency or take care that it is installed. The same strategy that of HuggingFace as well. They do not package PyTorch or Tensorflow or take care it is installed. Again this error arises from nasty imports folder, which is causing lot of concern. 
1. Do the tests required Fastai and PyTorch Lightning?
2. Do we need this in imports ?

Fixes possible.
1. I can install it while checking our wheel in package.yml file.
2. We can bundle this in requirements (bad idea limits the minimal installation).

Note that egg or wheel file will always install minimal version of mantisshrimp.

**Comment by lgvaz** [comment]
1. Tests requires fastai, lightning, efficientdet, albumentations, everything, so for now let's only use with [all]

2. We don't need it on imports =)

**Comment by oke-aditya** [comment]
Tests run on all. 
Wheel build check, checks our code only on our dependencies as imports.

**Comment by oke-aditya** [comment]
Okay. So the final broken things are Tests on Windows. This time it is not due to the distributed issue. It is due to refactor of COCO metrics. Windows we use different COCO Api repository. @lgvaz can you have a look it seems small dtype conversion error.

**Comment by oke-aditya** [comment]
Maybe this is the cause https://github.com/rbgirshick/py-faster-rcnn/issues/481. We are using an upgraded version of numpy but the coco API needs a lower one ? The COCO APi is always painful and a bottleneck.

**Comment by oke-aditya** [comment]
Currently we don't have any test to let us know if it is fixed. So I guess better will be to merge this and try to fix it up. It will be easier to fix as then we will know if Windows CI passes.

**Comment by oke-aditya** [comment]
Following tests fail only on Windows 
```
tests\metrics\test_coco_metric.py FF                            
tests\models\efficient_det\test_coco_metric.py F   
```

@lgvaz @ai-fast-track  as we said Windows support is experimental. And reason is PyCoco Tools and PyTorch both have its limitations.

Suggestions to Proceed: -
1. Merge this PR and try fixing up windows.
2. Skip these tests for now. Mark them as COCO in PyTest and skip it in windows.
3. Not support windows for now.

**Comment by ai-fast-track** [comment]
Does this (`pip install`) fix the issue with Windows when spinning the WIndows test machine?

## C.2- Installing cocoapi in Windows:
pycoco cannot be installed using the command above (see issue-185 in the cocoapi repository). We are using this workaround:

`pip install "git+https://github.com/philferriere/cocoapi.git#egg=pycocotools&subdirectory=PythonAPI"`

**Comment by oke-aditya** [comment]
I am using the same COCO API on workflow file. Still I'm getting error. You can have look at workflow [here](https://github.com/lgvaz/mantisshrimp/actions/runs/185688962/workflow)

**Comment by oke-aditya** [comment]
Bumping all the tests to PyTorch 1.6 with the new relase today,

**Comment by oke-aditya** [comment]
Wow. Additional trouble now. We can't bump to PyTorch 1.6. @lgvaz this needs multi-headed attention now (pun).

Problem is with Ross's EfficientDet tests. Also, a test of fastai is failing. 
Maybe raise an issue with Ross? 
Error is 
```
E       RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.
```
from 
``` preds = efficientdet.predict(model=fridge_efficientdet_model, batch=batch) ```

To avoid this from next time. We should run CI tests on PyTorch current as well as nightly. This was not very expected. But this increases our test count 2x. Which will now be too many.

oke-aditya:
K converting this to draft PR. Will raise 2 new PRs one for CI and one for the PyPI release.

oke-aditya:
Do let me know if we need to delete this branch

PR #236, opened 2020-07-28
Inference API

### LINKED ISSUE #193: Inference API
The real hard task here is how to create a clean api that is generic and not magical behind the scenes

## Step 1 - Create a Dataset
The images can have very different sources. We may already have records that were previously parsed, or the images might be in all in a folder, or the images might be all in memory as `np.array`

I think we can safely generalize all into two sources:
* Already parsed records
* Images in memory

For all other cases, you just need to get your images in memory. If they're in a folder, you open them, if they are urls, you download them. We don't want to provide functions for all these use cases, we let the user handle how to get the images in memory

After that, from both use cases you can create a Dataset and apply some transforms if needed. For a list of images in memory we can introduce a classmethod `Dataset.from_images`

Following this, all use cases boils down to `Dataset` , so we just need to think of a single solution for that.

### Proposal 1
Until now, I've been introducing a single method for prediction that works like so:

```python
batch = efficientdet.build_infer_batch(infer_ds)
preds = efficientdet.predict(model=model, batch=batch)
```

`build_infer_batch` takes the items from the `Dataset` and converts them into the format the model expects (and this is specific to each model). `preds` is a dictionary that looks similar for all models, e.g. if a model outputs bounding boxes, it will contain a key `bboxes` that will contain a list of `BBox` objects.

You may also want to predict in batches instead of everything at once, for that you can use a `DataLoader` like so:
```python
infer_dl = efficientdet.infer_dataloader(dataset=infer_ds, batch_size=4)

all_preds = []
for batch in infer_dl:
    preds = efficientdet.predict(model=model, batch=batch)
    all_preds.extend(preds)
```

While this is a bit verbose to write, it's also very simple and straight forward, I've been advocating against a method like `predict_dl` because that might confuse the user to when use one or another, but if we decide to go for it, it would look like so:

```python
infer_dl = efficientdet.infer_dataloader(dataset=infer_ds, batch_size=4)
preds = efficientdet.predict_dl(model=model, dl=infer_dl)
```

### Proposal 2
We can reduce boilerplate a little by internally calling `build_infer_batch` in `predict` and internally calling `infer_dataloader` in `predict_dl`:

```python
preds = efficientdet.predict(model=model, dataset=infer_ds)
preds = efficientdet.predict_dl(model=model, dataset=infer_ds, batch_size=4)
```

### Proposal 3
It can be argued (although I'm not sure) that the prediction using a `DataLoader` is generic enough to handle all use cases. In that sense, we could provide a single `predict` method that internally creates a dataloader:

```python
def predict(model, dataset, batch_tfms=None, **dataloader_kwargs):
    ...
```

## The challenging problem

I believe these proposals cover the task of generating predictions reasonably well. Let's say now we want to show our predictions, somehow we need to get the images out of the dataset that was used by the model, so we need to do the following:

```python
images = [sample['img'] for sample in infer_ds]
```

And then we can simply call `show_preds`:

```python
show_preds(images, preds)
```

The problem is that this approach would fail if you're using random transforms in the dataset (I know it's uncommon to use random transforms in inference, but that should not be a limitation), because each time you index the dataset you re-apply the transforms.

I tried playing with the idea of also returning the images on predict:
```
images, preds =  efficientdet.predict(model=model, dataset=infer_ds)
```

But that introduced another set of challenges and "magic" behind the scenes.

Another thing to keep in mind is that if you resized your images, all the preds will be related to that resized image, you would need to manually rescale the predictions to match your original image.

* Can we somehow return preds related to the original images?
* How can we link the transformed images to the predictions, in a way we don't need to reindex the `Dataset`?

lgvaz:
closes #193 

Prediction for all models follow the standard api:

```python
batch, samples = faster_rcnn.build_infer_batch(infer_ds)
preds = faster_rcnn.predict(model=model, batch=batch)
```

`batch` is dependent on what model you're using, it's a tuple of input tensors in the format the underlying model expects.

`samples` is common for all models, it's a list of dicts containing the image, possibly targets (bboxes, masks), all python primitive types and numpy arrays.

A new method `predict_dl` was also introduced.

For more info take a look at `examples/inference.py`


ai-fast-track:
@lgvaz It looks good for me.

oke-aditya:
LGTM. Clean API. I hope it will remove some of the Pycoco tools issue too.

Issue #239, opened 2020-07-28
Update the getting_started.ipynb tutorial


ai-fast-track:
## 🐛 Bug
The getting_started.ipynb tutorial need to be updated because of the changes in both the models and the inference API.

lgvaz:
We also need to update the training code snippet that shows on the readme

FraPochetti:
I am also stumbling upon the below issue.
I navigated through the source code and indeed I don't seem to find the `_root` attribute defined anywhere.

FYI, I am trying to peel and the onion and figure out how `Pets(Voc)XmlParser` works as I'd like to parse [this](https://www.kaggle.com/mbkinaci/fruit-images-for-object-detection) very similar toy dataset with it.

![image](https://user-images.githubusercontent.com/1206519/88794198-60f61780-d19e-11ea-9149-c809c4400790.png)

FraPochetti:
>  I don't seem to find the _root attribute defined anywhere

Found it [here](https://github.com/FraPochetti/mantisshrimp/blob/3b2084ef1901d474210af4b1f9ad949550c8c8f6/mantisshrimp/datasets/voc/parsers.py#L33) :).
I'll keep digging from there.

lgvaz:
@FraPochetti That's correct! Sorry for the outdated examples, the API is moving very very fast and it's hard to keep up hahah

#241 updates the quickstart on the readme, it's the latest news

PR #241, opened 2020-07-29
updates training example

lgvaz:
Related to #239 

@ai-fast-track, can you take a look?

I'm not used with our new documentation procedure, do I have to modify any other file?


lgvaz:
We need to modify black params to only check files inside mantisshrimp folder

Issue #246, opened 2020-07-31
How to install efficientdet?


lgvaz:
None of the pypi installation methods will install efficientdet, the user has to manually install it via `pip install git+...`, but this is not mentioned anywhere in the documentation

Since it's in the core of our library to have soft dependencies, maybe it's a good idea to have a section in the docs explaining what you have to install if you want to use X. Say in the near future we add support for yolo, we need a section explaining the steps to install it (or point to the original installation instructions). This goes well with #245 

@oke-aditya @paras-jain @ai-fast-track , what are your thoughts?

oke-aditya:
This is not exactly soft dependency as model definition is very important to our package, we cannot let it go away from the package.
We can add this to our repo only if we get `.egg` file from the repositories. Also, we need to check what dependency these repositories use, if they are minimal we can package them.

lgvaz:
It's a soft dependency, mantis does not fail to import if you don't have efficientdet

Let's say you only want to train faster_rcnn, you shouldn't have to install efficientdet for that. And this should be true for every new model that we add

oke-aditya:
What if we get `.egg` file from this repo after a few months. Can we bundle it together ?

lgvaz:
Getting the `.egg` files just seem to hacky, we don't want to deal with that

It's actually easier to have instructions on how to install the package

oke-aditya:
The problem is when we ship for release 0.1. The package from github repo may get updated and hence incompatibly with release. Suppose we lock our dependencies when it was working fine, then we receive an update from these packages that break stuff.
So it is safer to ship it with the corresponding `.egg` file.

lgvaz:
The efficientdet dependency will also be locked to a specific version on our release, why are updates going to break stuff?

oke-aditya:
How will we lock efficient det version ? Do we lock to some commit stamp?

lgvaz:
We lock a commit stamp, yes

oke-aditya:
Can we build a wheel for that commit stamp and provide as assets with our repo ?

lgvaz:
My answer is: I don't know, and it should not be our priority now, not for 0.1.0

I'll start a 0.2.0 roadmap and add this there

lgvaz:
Let's stick with adding installation instructions for efficient for this release, I'll close this one once we do that

lgvaz:
Efficient is now installed with `[all]` =)

PR #265, opened 2020-08-04
Updates dependencies

lgvaz:
relates to #246


oke-aditya:
Are you sure we want to bundle PyTorch in our dependencies? Users can install that from PyTorch website. I feel we can leave that to them. Sometimes PyTorch does not install properly. It fetches wrong version of PyTorch with torchvision. And thus causes problems.

lgvaz:
We should still mention in the docs that they should install from the official instructions

I actually agree with what you're saying, the issue is that fastai/lightning installs pytorch anyways... lol

Issue #287, opened 2020-08-08
parsers namespace


lgvaz:
Instead of exporting everything from parser into the default namespace, we can export everything into a `parsers` namespace. For example, the following:

```python
class WheatParser(DefaultImageInfoParser, FasterRCNNParser):
    pass
```

would change to:
```python
class WheatParser(parsers.DefaultImageInfo, parsers.FasterRCNN):
    pass
```

The main idea is to facilate the life of the user with autocompletion, so, for everything related to parsers one can just do `parsers.<autocomplete>`. 

You can see this trend is already happening in other parts of the library, we have `datasets`, `tfms`, and the same happens for models, `faster_rcnn`, `efficientdet`.

I think this makes the life of especially those who just joined the library much easier.

---

Another reason is that the COCO parser is going to be moved from `datasets` to `parsers`, it's much more intuitive to do:

```python
parsers.coco
```

then:
```python
coco_parser
```

ai-fast-track:
Good idea: it standardizes the way we use the library.

PR #288, opened 2020-08-08
moves coco and voc parser from datasets to parsers

lgvaz:


lgvaz:
closes #287

ai-fast-track:
It looks good to me: All standard parsers are now under the `parsers` folder. All the other parsers are located in the corresponding datasets.

PR #298, opened 2020-08-11
fixes error with mask training

lgvaz:


examples/birds.py
```diff
@@ -0,0 +1,126 @@
+from mantisshrimp.all import *
+import scipy.io
```
oke-aditya, line 2: Does this add extra scipy dependency. I guess matplotlib already has that.

lgvaz: This is only inside examples though =)

PR #368, opened 2020-09-02
implements draw methods with opencv

lgvaz:


lgvaz:
@ai-fast-track the only function we should need to change to fix the ugly labels is `_draw_label`

oke-aditya:
Can this be done later switched with native torchvision as they are going to implement drawing boxes etc.

lgvaz:
@oke-aditya Even if they start supporting it natively it's something we want to control ourselves.

There is no reason to depend on their code for this.

oke-aditya:
Right. Does this makes open-cv an additional dependency ?

I guess a soft one though

lgvaz:
open-cv was already a dependency =), this can make matplotlib optional now heh

PR #399, opened 2020-09-10
Fixing custom domain issue

ai-fast-track:
I created a CNAME file in the root folder. That file name is also copied in the docs folder when the docs are built.


ai-fast-track:
Closed: Fixed in the PR #399

PR #444, opened 2020-09-24
Fix aug_tfms docs link

davanstrien:
The current link in the docs to `aug_tfms` is broken. I assume this should point to the docs page for `aug_tfm` and have updated link to point to this.


ai-fast-track:
Thank you very much @davanstrien  😃

PR #454, opened 2020-09-26
Port to official Fastai v2

bguan:
Global search and replace "fastai2" to "fastai".
Fix test failures, import issues.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,24 +1,24 @@
 name: ice
 channels:
-- fastai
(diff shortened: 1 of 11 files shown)


lgvaz (approved):
Very very nice!! This is immensely helpful, thanks a lot!!

Looks good to me, let's test it out on the master branch for a few days, and then we can make a release from it =)

Welcome to the icevision contributors club 😄

PR #481, opened 2020-10-17
Added wandb support

ai-fast-track:
I added wandb support. Creating wand images with both predicted bboxes and ground truth bboxes (optional using `add_ground_truth` argument) is one line:

```python
wandb_images = wandb_img_preds(samples, preds, class_map, add_ground_truth=True)
```

Logging in wandb:
```python 
wandb.log({"Predicted images": wandb_images})
```


lgvaz (commented):
We still have to further discuss how we want to add wandb to the core library but it's a better idea to make it a soft dependency

PR #510, opened 2020-10-27
Fixes autofix mask

lgvaz:
TODO:
- [x] RLE can default to coco style
- [x] `RLE.to_mask` can use pycocotools (time both approaches and stick with the fastest): 
  - Does it matter? Because everything gets converted to erle
- [x] Refactors names 
  - `MaskArray`
  - `EncodedRLEs` method and attribute names
- [x] Refactor `MaskArray.from_masks`
- [x] Figure out if `convert_coco_records` is working with encoded RLEs


lgvaz:
image

PR #515, opened 2020-10-29
Update CONTRIBUTING.md

FraPochetti:
The `black .` command should be invoked before `git add` in order for the reformatting to properly be committed and pushed.


lgvaz (approved):
Good catch!!

lgvaz:
Ow, just now realized I forgot to merge hahaha

PR #516, opened 2020-10-29
Added pretrained argument to Mask-RCNN model definition

ai-fast-track:
Added pretrained argument to Mask-RCNN model definition to avoid systematically loading the pretrained weights. This covers the scenario where we already have another pretrained weights such the PennFudan model.


lgvaz (changes requested):
Hi Farid!

Tricky thing here is that this argument only takes effect when `backbone==True`, before we merge, can you add the parameter to the docstring and clarify that?

Also, can you make the same change on `faster_rcnn`?

lgvaz (approved):
Perfect! Amazing description of the `pretrained` argument on the docstrings

Issue #490, opened 2020-10-19
Wandb + fastai + RCNN error


lgvaz:
## 🐛 Bug
**Describe the bug**
An error is raised (strack trace below) when training a rcnn model with fastai and wandb.

**To Reproduce**
Train a RCNN model with fastai and the WandbCallback.

**Expected behavior**

**Why is this happening**
I believe the `RCNNCallback` and the `WandbCallback` are in conflict somehow. This error does not happen if the `WandbCallback` is used or if `efficientdet` is used instead of rcnn.

**Stack Trace**
```python
/usr/local/lib/python3.6/dist-packages/torch/nn/modules/module.py in _call_impl(self, *input, **kwargs)
    734                 else:
--> 735                     var = var[0]
    736             grad_fn = var.grad_fn

TypeError: 'ImageList' object does not support indexing
...

During handling of the above exception, another exception occurred:

/usr/local/lib/python3.6/dist-packages/fastcore/foundation.py in __getattr__(self, k)
    157         if self._component_attr_filter(k):
    158             attr = getattr(self,self._default,None)
--> 159             if attr is not None: return getattr(attr,k)
    160         raise AttributeError(k)
    161     def __dir__(self): return custom_dir(self,self._dir())

AttributeError: 'Learner' object has no attribute 'smooth_loss'
```

FraPochetti:
Is it ok if I look into this one as well (before diving together into KeyPoint detection)?

lgvaz:
Perfect! I assigned it to both of us, I think this might be a tricky one to resolve (hopefully it's easy hahahah).

I have some ideas about what is happening, we can talk about it tomorrow as well, but as I said in the issue, it's about how both callbacks are interacting with each other

FraPochetti:
I will make sure to take a look at it before tomorrow's chat ;)

lgvaz:
This error might be instead being caused by replacing `learn.loss` attribute in `models/rcnn/fastai/learner.py`:

```python
    # HACK: patch AvgLoss (in original, find_bs gives errors)
    class RCNNAvgLoss(fastai.AvgLoss):
        def accumulate(self, learn):
            bs = len(learn.yb)
            self.total += fastai.to_detach(learn.loss.mean()) * bs
            self.count += bs

    recorder = [cb for cb in learn.cbs if isinstance(cb, fastai.Recorder)][0]
    recorder.loss = RCNNAvgLoss()
```

FraPochetti:
[...]

PR #524, opened 2020-11-03
monkey-patching wandb.watch when training a fastai RCNN learner with …

FraPochetti:
…the WandbCallback

This one addresses issue #490

### DIFF
--- a/icevision/models/rcnn/fastai/learner.py
+++ b/icevision/models/rcnn/fastai/learner.py
@@ -6,11 +6,16 @@ from icevision.models.rcnn.loss_fn import loss_fn
 from icevision.models.rcnn.fastai.callbacks import *
 
 
+def noop_watch(models, criterion=None, log="gradients", log_freq=1000, idx=None):
+    print("inside noop_watch")
+    return []
+
+
 def rcnn_learner(
     dls: List[Union[DataLoader, fastai.DataLoader]],
     model: nn.Module,
     cbs=None,
-    **kwargs
+    **kwargs,
 ):
     learn = adapted_fastai_learner(
         dls=dls,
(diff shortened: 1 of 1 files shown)


lgvaz (commented):
Looks good! Just a question before we merge this, if instead of patching we use `log=None`, is the bug solved?

FraPochetti:
> if instead of patching we use log=None, is the bug solved?

Nope. That just doesn't log the gradients.

lgvaz:
Okay, makes a lot of sense!!

So, the final thing before we merge, can you add a `logger.warning` if `if len(is_wandb) == 1:`?

It would be nice to open a new issue in icevision to show your findings, you can just copy and paste the messages you sent me on discord, explaining that torchvision rcnn models are having a problem with the wandb hooks and the best quick solution we found is to simply don't add the hooks

Then in the warning message you could say something like "Wandb quickfix implemented, for more info check issue #number_of_issue"

FraPochetti:
Sure will do!
Shall I just document my findings [here](https://github.com/airctic/icevision/issues/490) or create a new issue from scratch?

FraPochetti:
I added a final comment to [this](https://github.com/airctic/icevision/issues/490) issue describing what we did.
Is this a sufficient level of detail?

lgvaz:
The level of detail is perfect!

Open another issue because the original one contains useless information and will be closed by this PR. It's easier to have a direct and clear issue for further reference =)

FraPochetti:
Done.
[Issue](https://github.com/airctic/icevision/issues/527) created and PR updated.

lgvaz (approved):
Looks perfect! Thanks SO MUCH again Francesco, as I said before, the amount of line changes doesn't reflect at all the work put into this :heart:

PR #529, opened 2020-11-04
adding keypoints annotations for 2 sample images

FraPochetti:


FraPochetti:
Btw, I am not getting the below locally.
Everything is passing on my machine.
image

lgvaz:
The error is not related to your PR, it's happening on master actually 😅 

I'll fix it asap

lgvaz (commented):
Look very good!!

I just have one question on `draw_keypoints`, currently we're using `kps.human_conns` to draw the connections, any plans on how to support arbitrary keypoints?

We can merge this first part if you think it's a good idea, do you want to me to squash all commits into one or do you want to try rebasing and cleaning the commits? Rebase and re-structuring the commits is a very helpfull functionality that I just learned recently, I'm more than happy to hop into a call to help you do that if needed :smiley:

FraPochetti:
> any plans on how to support arbitrary keypoints?

This is a very good point. First of all, this is coming from COCO. I think the order of the connections should stay the same across datasets, as long as we are talking about humans. By that, you mean non-human keypoints?

> We can merge this first part if you think it's a good idea

I think it is. Anything against that?

> do you want to try rebasing and cleaning the commits?

I will try to figure it out myself man! It seems indeed a very nice functionality.

lgvaz:
> By that, you mean non-human keypoints?

Yeah yeah

> I think it is. Anything against that?

Not really, the only downside I see is that the commits regarding keypoints implementation are not going to stay next to each other. But that is not an issue at all.

 > I will try to figure it out myself man! It seems indeed a very nice functionality.

It's amazing indeed! It will give you a lot of freedom with git actually

FraPochetti:
I haven't stumbled upon non human Keypoints yet. Have you?

lgvaz:
We did, for example that head pose estimation we have on fastai, or facial keypoints or even [cat facial keypoints](https://www.kaggle.com/crawford/cat-dataset) :laughing:

FraPochetti:
Ah! Well, then let me try to spot some of those `non-human` examples from COCO, if there are any, and I will adapt the code accordingly. 
I guess I will have to check for the class, e.g. `if class=="human": self.connections = human_connections` etc. As of now, there is a bug, as I assume keypoints are only from humans. 
Do you know if COCO contains any such images?
I will look for them and add them to our samples.

FraPochetti:
So, not quite ready for merging!

lgvaz:
> Well, then let me try to spot some of those non-human examples from COCO

Oh, sorry some miscommunication on my part, for COCO I think we only have human keypoints. But the `keypoints` class and the visualization should not be specific to COCO, maybe we should discuss further how to handle such cases.

Maybe one of the arguments you pass to `keypoints` are how to connect the points?

Or maybe you have to subclass `keypoints` to change how the points are connected?

I'm not sure what is the best strategy, just throwing a bunch of ideas here :smile:

FraPochetti:
>  But the keypoints class and the visualization should not be specific to COCO

Totally agree.
Not sure about the best strategy either actually.
I am just thinking about how this would work in practice, in an automated fashion, without the user "telling" the `keypoints` class which object those keypoints belong to.
Probably the `keypoints` class needs to be aware of the class of the object, as I was proposing.
Not sure though.

lgvaz:
Maybe we can have something like:

```python
class KeypointsInfo:
    def __init__(self, names, connections):
        ...

COCO_HUMAN_KEYPOINTS_INFO = KeypointsInfo(['nose', ...], ...)

class KeyPoints:
    def __init__(self, keypoints: Union[List[int], np.array], keypoints_info:KeypointsInfo=COCO_HUMAN_KEYPOINTS_INFO):
        ...

```

FraPochetti:
I like the solutions but this is just assuming again that, by default, Keypoints are from a human. What if, within the same image, we have a human and a cat, both with Keypoints provided? How do we know how to instantiate 2 different classes accordingly?

FraPochetti:
> for example that head pose estimation we have on fastai, or facial keypoints

btw, those should be a subset of the 17 keypoints currently supported, so the functionality should be the same.
For anything that is not-human the issue persists.

FraPochetti:
Another option is that we disable connections by default.
If you think about it, this whole issue is related to how to connect the points, hence it is just a visualization thing.
Nothing to do with the modeling. 
So, given we are not sure what is represented on the image, by default we don't plot the connections.
We could add a `add_connections` argument to `show_sample` or something like that to give the user the possibility to add those.
This would ensure that, if we have a person and a cat in the same picture, the dots are plotted and that's it. Nothing else to handle.

Technically, the "connection" info should be contained in the dataset, and, in theory, we could try parsing it.
In COCO, this piece of data is inside the `annotation.json` within `categories --> skeleton` (one skeleton per class, e.g. one for human, one potentially for cat, etc).
I am not extracting it from COCO. I added the human skeleton by default to the class, but we could try parsing it each time a parser is instantiated.

lgvaz:
> What if, within the same image, we have a human and a cat, both with Keypoints provided?

Excellent question... I think this also hints that passing `add_connections` to `show_sample` would not be ideal, since for `show_samples` we would need to pass a list of connections adding a lot of overhead to the user. We should also try to maintain the show methods agnostic and in the future we can refactor it in a way that each data class knows how to show itself (effectively eliminating the need for the if statements inside `show_sample` (`if mask in sample:...`)

 > Technically, the "connection" info should be contained in the dataset, and, in theory, we could try parsing it.

Yeah, this is a good insight.

I'm still having a problem to grasp the correct approach here, we need a bit more time and information. Since this is only a visualization problem, what do you think of the idea to just move forward in implementing the model? This could give us some additional insight that might help us solve the current issue

FraPochetti:
Yes, I agree.
Shall we disable the connections for the time being?
I mean, I will just comment the following code in `draw_keypoints`, which is responsible for drawing connections.
```
    for sk in sks:
        if np.all(v[sk] > 0):
            cv2.line(
                img,
                (x[sk][0], y[sk][0]),
                (x[sk][1], y[sk][1]),
                color=color,
                thickness=3,
            )
```
This way the code is kept there and we can switch it on back when we figure this out inside the `KeyPoints` class.
Wdyt?

Then I rebase, we merge and we move to datasets/dataloaders.

FraPochetti:
Side note, can `albumentations` handle `keypoints` augmentations?

FraPochetti:
Also, I guess at this point we would need an actual dataset to play around with.
Shall we go with BIWI?

FraPochetti:
I got excited and started working on the `Dataset`. We'll merge another time! :D

lgvaz:
> Shall we disable the connections for the time being?

Yeah! Commenting it out is fine

> Shall we go with BIWI or this one?

I would suggest working with two datasets, an easy one (probably BIWI?) and a harder one(the other one you suggested, or the cats one?). Do all the development with the easy one and when you think it's working check with the harder one.

> I got excited and started working on the Dataset. We'll merge another time! :D

Cool cool!

lgvaz (approved):
The latest changes look good!

I tested the functionality on [this](https://colab.research.google.com/drive/1R_3ZFktvy8pLmbRn5SESwXX8fwVYPDFI?usp=sharing) colab notebook and it's working as expected.

Before we merge this, can you use rebase to clean the commits? For example, `added`, `added draw_predictions and relevant tests` and `Merge branch 'keypoints' of https://github.com/FraPochetti/mantisshrimp…` can all be combined into a single commit.

I'm asking for this because I want to use the `rebase` option while merging so all of the changes doesn't get condensed into a single commit.

The command you'll need to run is `git rebase -i 7a929941f7ede53b4b82595d08aa2240dad5454d` then `fixed` and `squash` specific commits and then force push.

I can also do this cleaning on my side if you don't want to dig into it

PR #537, opened 2020-11-10
fixing visualization after augmentations

FraPochetti:


icevision/models/torchvision_models/faster_rcnn/dataloaders.py
```diff
@@ -85,10 +85,14 @@ def _build_train_sample(
     if len(record["labels"]) == 0:
```
lgvaz, line 85: Instead of modifying the faster_rcnn dataloader we should create a specific extended dataloader for keypoints_rcnn, similarly to what is done for mask_rcnn dataloaders

icevision/models/torchvision_models/faster_rcnn/prediction.py
```diff
@@ -61,9 +65,25 @@ def convert_raw_prediction(raw_pred: dict, detection_threshold: float):
...
 
-    return {
+    d = {
         "labels": labels,
         "scores": scores,
         "bboxes": bboxes,
         "above_threshold": above_threshold,
     }
+
+    if raw_pred.get("keypoints") is not None:
```
lgvaz, line 75: Here we should also implement the specific functionality into `keypoint_rcnn/prediction.py`, in that way we also don't need the `if raw_pred.get("keypoints") is not None:`

lgvaz (changes requested):
Hi! 

First of all, amazing job Francesco! Keypoints from 0 to 1 you implemented :tada: 

There are a lot of files in this one, I reviewed some of them today and I'll do the others tomorrow =)

icevision/models/torchvision_models/faster_rcnn/dataloaders.py
FraPochetti: Done!

icevision/models/torchvision_models/faster_rcnn/prediction.py
FraPochetti: Done! 
Also added a test on predictions, which I did not realize was missing.

lgvaz (changes requested):
Alrighty, I checked all files but the albumentations related ones, for that I'll check locally on my editor, there are a lot of things happening there =)

icevision/models/torchvision_models/keypoint_rcnn/prediction.py
```diff
@@ -0,0 +1,74 @@
...
+        raw_pred=raw_pred, detection_threshold=detection_threshold
+    )
+
+    above_threshold = preds["above_threshold"]
+    kps = raw_pred["keypoints"][above_threshold]
+    keypoints = []
+    for k in kps:
+        k = k.cpu().numpy()
+        k = list(chain.from_iterable(k))
+        if sum(k) > 0:
```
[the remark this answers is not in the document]
FraPochetti: Sure

icevision/visualize/draw_data.py
```diff
@@ -243,8 +243,24 @@ def draw_keypoints(
     #             color=color,
     #             thickness=3,
     #         )
+    if np.count_nonzero(y) > 0 and np.count_nonzero(x) > 0:
```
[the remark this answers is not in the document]
FraPochetti: Yes. This is technically useless as the logic `sum(k) > 0` filters `KeyPoints` objects upstream, before they even get created.
`draw_keypoints` was one of the first functions I put together. The `sum(k) > 0` logic came later on.

icevision/visualize/draw_data.py
```diff
@@ -243,8 +243,24 @@ def draw_keypoints(
...
+        min_x, max_x = x[x > 0].min(), x[x > 0].max()
+        min_y, max_y = y[y > 0].min(), y[y > 0].max()
+
+        Rfactor = math.sqrt((max_x - min_x) * (max_y - min_y)) / math.sqrt(
```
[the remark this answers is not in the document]
FraPochetti: I shamelessly copied it from [here](https://github.com/liruilong940607/OCHumanApi/blob/958aa2046ba4f5760fcc94cb458da8fe72cedf4f/ochumanApi/vis.py#L92).
I forgot to add a reference.

FraPochetti:
Thanks a lot for going through all of this man! Super appreciate your help!

icevision/models/torchvision_models/keypoint_rcnn/prediction.py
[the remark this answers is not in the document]
FraPochetti: Done

icevision/visualize/draw_data.py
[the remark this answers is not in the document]
FraPochetti: Done

icevision/tfms/albumentations/tfms.py
```diff
@@ -111,41 +113,170 @@ def apply(
...
+                for i in range(0, len(l), len(l) // len(keypoints))
+            ]
+            assert len(l) == len(keypoints)
+            cl = keypoints[0].labels
+            # `if sum(k) > 0` prevents empty `KeyPoints` objects to be instantiated.
+            # E.g. `k = [0, 0, 0, 0, 0, 0]` is a flattened list of 2 points `(0, 0, 0)` and `(0, 0, 0)`. We don't want a `KeyPoints` object to be created on top of this list.
+            out["keypoints"] = [KeyPoints.from_xyv(k, cl) for k in l if sum(k) > 0]
         if labels is not None:
             out["labels"] = [labels[i] for i in d["labels"]]
+            if keypoints is not None:
```
lgvaz, line 183: How can we refactor this so we don't need to do the check `if keypoints is not None:` in each object?

The logic that happens to filter the objects seems repetitive: `d["masks"][i] for i, k in zip(d["labels"], l) if sum(k) > 0`

Is it possible to somehow write a function that does this filtering and then call this function irrespective if `keypoints is None`?

lgvaz (commented):
Alrighty, last step is the transforms, for sure the most complicated logic of all of this

I tried thinking of other approaches to handle all that is happening there but I couldn't think of anything simpler, so let's just go with this and refactor what we can =)

icevision/tfms/albumentations/tfms.py
FraPochetti: Yep makes sense. I will try to do that.

FraPochetti: I refactored the logic a bit. Is this sufficient?

FraPochetti (commented):
All good for me. Much neater code!

icevision/tfms/albumentations/tfms.py
lgvaz: Resolving because it was addressed on discord =)

lgvaz (approved):
That's it Francesco! Thanks **a lot** for this awesome contribution!! :heart: :heart: 

Here we go!

FraPochetti:
This is fantastic!

PR #538, opened 2020-11-11
Retinanet

### LINKED ISSUE #145: Add RetinaNet with custom Backbones and helpers.
## 🚀 Feature

Retina net is a SOTA model for object detection. Torchvision models is soon going to have implementation of it very soon [here](https://github.com/pytorch/vision/pull/1697) Which is very similar to faster RCNN. 
We have support for lot of backbones with fasterRCNN same can be extended here too. 
Also focal loss is an additional plus too. 
Plus the fastai and lightning adapters almost stay the same. 

Just keeping this open so that we would immediately have this feature when it comes.

lgvaz:
closes #145


FraPochetti:
This is great work man!

lgvaz:
> This is great work man!
@FraPochetti Thank you! This one was easier than expected huh? hahahah

FraPochetti:
Are you merging into master?

lgvaz:
Soon enough, just need to finish some additional tests

lgvaz:
Will this cause conflicts with your work?

FraPochetti:
I don't think so

PR #539, opened 2020-11-11
Added comments that EfficientDet requires the input to be divisible by 128.

### LINKED ISSUE #535: EfficientDet images_size has to be divisible by 128
## 📓 Documentation Update
**What part of documentation was unclear or wrong?**
It has to be clear that the image size for the efficientdet model has to be divisible by 128

**Describe the solution you'd like**
Add this information to the `efficientdet/model.py` docstring and also make it explicit in the tutorials that use efficientdet.

fstroth:
This should resolve #535 .

### DIFF
--- a/examples/efficientdet_pets_exp.md
+++ b/examples/efficientdet_pets_exp.md
@@ -108,7 +108,7 @@ In this example, we show how to train an EffecientDet model on the PETS dataset 
     # Datasets
     # Transforms
     presize = 512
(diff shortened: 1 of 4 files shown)


lgvaz (changes requested):
Looks very good! Can you also add this info to `notebooks/quickstart.ipynb`?

lgvaz (approved):
That's it, awesome!

Thank you for the PR :heart:

PR #541, opened 2020-11-13
Issue 534 : Add precommit

partham16:
- `flake8` has an `--exit-zero` in CI which is currently commented out in `.flake8` - as they're incompatible 
- minor update in `.gitignore` : to exclude venv like `pythonenv3.8`

*btw, is it preferable to squash the 3 commits into a single one?*

##### the documentation needs to be upgraded as well, but, let's have that after this gets reviewed/merged

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,7 @@
+[flake8]
+ignore = E203, E266, E501, W503, F403, F401, F405
+# from ci-all-testing.yaml
+select = E9,F63,F7,F82
+# `--exit-zero` can't be set in config
+# max-line-length = 127
+# max-complexity = 10
--- a/.flake8
+++ b/.flake8
@@ -2,6 +2,6 @@
 ignore = E203, E266, E501, W503, F403, F401, F405
 # from ci-all-testing.yaml
 select = E9,F63,F7,F82
+max-line-length = 127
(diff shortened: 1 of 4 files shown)


.flake8
```diff
@@ -0,0 +1,7 @@
+[flake8]
+ignore = E203, E266, E501, W503, F403, F401, F405
+# from ci-all-testing.yaml
+select = E9,F63,F7,F82
+# `--exit-zero` can't be set in config
+# max-line-length = 127
```
lgvaz, line 6: Should max-line-lenght be uncommented?

lgvaz (commented):
Looks very good!

I just made a simple question before we merge

lgvaz:
> btw, is it preferable to squash the 3 commits into a single one?

yeah, I will do that when merging =)

.flake8
partham16: I added both that, and `max-complexity` - but, then `--exit-zero` part won't work there!

But, I guess as we have black already, that shouldn't be a concern!

partham16:
`max-line-length` isn't strictly necessary, as `E501` ignores the line-length error anyway!

lgvaz (approved):
Amazing!! Very very clean and concise!

Thanks for the PR ❤️

PR #550, opened 2020-11-15
Closes 549 : Update coco_parser - ensure img_dir is `Path`

partham16:
updated to `self.img_dir = Path(img_dir)`
otherwise `self.img_dir / ...` in `def filepath` gives
 `TypeError : unsupported operand type(s) for /: 'str' and 'str'`


FraPochetti:
Great!

lgvaz (approved):
NICE! Looks good!!

Thanks a lot for finding and fixing this bug :rocket: :heart:

PR #564, opened 2020-11-25
Added ResNeSt backbones support

### LINKED ISSUE #561: Add ResNeSt backbones
## 🚀 Feature
ResNeSt aims to replace a ResNet backbones in both classification and object detection tasks and substantially improve accuracy.
It would be beneficial for IceVision to support the different ResNeSt architectures.

Paper: ResNeSt: Split-Attention Networks - https://arxiv.org/pdf/2004.08955.pdf 

Discord discussion: https://discordapp.com/channels/735877944085446747/735877944517591151/780303631738994700

ai-fast-track:
I created a separate module call `resnest_fpn` under `icevision.backbones` where I added the `resnest_fpn_backbone` that creates (pretrained) resnest backbones for each of the 4 architectures: `resnest50, resnest101, resnest200, resnest269`

I use the same `patch_param_groups(model)` de fined in the `resnet_fpn` module.

closes #561


icevision/backbones/resnest_fpn.py
```diff
@@ -0,0 +1,120 @@
...
+        return_layers,
+        in_channels_list,
+        out_channels,
+        extra_blocks=extra_blocks,
+    )
+
+
+def _resnest_fpn(name, pretrained: bool = True, **kwargs):
+    model = resnest_fpn_backbone(resnest_name=name, pretrained=pretrained, **kwargs)
+    patch_param_groups(model)
```
lgvaz, line 102: so after these changes the same function to patch parameters on resnet_fpn also works on resnest? cool!

icevision/backbones/resnest_fpn.py
```diff
@@ -0,0 +1,120 @@
...
+    Arguments:
+        backbone_name (string): resnet architecture. Possible values are 'ResNet', 'resnet18', 'resnet34', 'resnet50',
+             'resnet101', 'resnet152', 'resnext50_32x4d', 'resnext101_32x8d', 'wide_resnet50_2', 'wide_resnet101_2'
+        norm_layer (torchvision.ops): it is recommended to use the default value. For details visit:
+            (https://github.com/facebookresearch/maskrcnn-benchmark/issues/267)
+        pretrained (bool): If True, returns a model with backbone pre-trained on Imagenet
+        trainable_layers (int): number of trainable (not frozen) resnet layers starting from final block.
+            Valid values are between 0 and 5, with 5 meaning all backbone layers are trainable.
+    """
+    backbone = resnest_name(pretrained=pretrained)
```
lgvaz, line 64: I guess this is the only line different from the pytorch implementation? The problem was that they were using `module.__dict__` to get the model right?

icevision/backbones/resnest_fpn.py
```diff
@@ -0,0 +1,120 @@
...
+
+from icevision.imports import *
+from icevision.utils import *
+from icevision.backbones.resnet_fpn import patch_param_groups
+from torchvision.ops.feature_pyramid_network import LastLevelMaxPool
+from torchvision.models.detection.backbone_utils import (
+    resnet_fpn_backbone,
+    BackboneWithFPN,
+)
+import resnest.torch
```
lgvaz, line 16: We should make this a soft import, should I work on it?

lgvaz (changes requested):
Hi! Some suggestions =)

icevision/backbones/resnest_fpn.py
ai-fast-track: Yes, we are even using the same `patch_param_groups()` used by the `resnet_fpn`

icevision/backbones/resnest_fpn.py
ai-fast-track: Yes. The other small difference is that we are not using this argument `norm_layer=misc_nn_ops.FrozenBatchNorm2d`. We know that choice is debatable given the fact that we usually train the BatchNorm layer. In the case of the `resnet_fpn` they pass it to the resnet method like this:
```
backbone = resnet.__dict__[backbone_name](
        pretrained=pretrained,
        norm_layer=norm_layer)
```

icevision/backbones/resnest_fpn.py
ai-fast-track: That would be great!

PR #568, opened 2020-12-02
Update parser.py

fstroth:
Added comment to explain the usage of line 111: Record = self.record_class()


lgvaz:
I added a single commit to fix the formatting issue! The comment looks very good

Thanks for the PR!

Issue #563, opened 2020-11-25
Plot Top Losses


RibenaMapleSyrup:
## 🚀 Feature
**Is your feature request related to a problem? Please describe.**
I would like to plot top losses for predictions made by a trained model. This would be really useful to help me figure out where my trained model is making mistakes. With show_results I can view a model's predictions however loss would also be useful as loss encompasses how confident the model is. With this additional information I could better clean my dataset and work towards improving accuracy. 

**Describe the solution you'd like**
I would like to write a function that accepts samples and predictions as input parameters and outputs an ordered dictionary pairing samples with loss values. 

**Describe alternatives you've considered**

**Additional context**
It doesn't seem possible to calculate loss when model is in evaluation mode.

FraPochetti:
Hey @lgvaz, can you please assign me to this one?

FraPochetti:
Feature implemented [here](https://github.com/airctic/icevision/pull/589).
Closing.

RibenaMapleSyrup:
Awesome work!! Look forward to working with this feature. Thanks so much :D

PR #589, opened 2020-12-13
adding plot_top_losses for faster_rcnn

FraPochetti:
addressing #563 
extensive discussion already happening on Discord.


FraPochetti:
@lgvaz any specific feedback so far?
This seems a good start to me.
Example usage:
```
samples_plus_losses, preds, losses_stats = faster_rcnn.plot_top_losses(model=model, dataset=valid_ds, sort_by="loss_total", n_samples=6)
```
Result:
image

FraPochetti:
Actually let me create a Colab illustrating the feature. I think it will be also easier to provide feedback. Will do that tonight

FraPochetti:
Here the colab: https://colab.research.google.com/drive/1lAtBMVWD7hb0EUK4exjK2aRCGxTHIA9u?usp=sharing

icevision/models/torchvision_models/faster_rcnn/show_results.py
```diff
@@ -31,3 +36,103 @@ def show_results(
...
+    dataset: Dataset
+
+    Returns
+    -------
+    samples_plus_losses: those are the samples (dictionaries) contained in the original dataset
+                          enriched with losses.
+    losses_stats: dictionary containing losses stats (min, max, mean, 1-25-50-75-99 quantiles)
+                  for each one of the losses.
+    """
+    device = torch.device("cpu")
```
lgvaz, line 62: Can we also run this on GPU?

icevision/models/torchvision_models/faster_rcnn/show_results.py
```diff
@@ -31,3 +36,103 @@ def show_results(
...
+    model = model.to(device)
+    dl = valid_dl(dataset, batch_size=1, num_workers=0, shuffle=False)
+
+    torch.manual_seed(0)
```
lgvaz, line 66: If we want all samples to have the same dropout values and all, should we move the manual seed to inside the for loop?

Actually, should we try to find a way to put the model in eval model while still having it to calculate the loss?

Inside torchvision in `generalized_rcnn.py` we see that the forward loop checks for the variable `self.training`, what happens if we do:
```python
model.eval()
model.training = True
```

icevision/models/torchvision_models/faster_rcnn/show_results.py
```diff
@@ -31,3 +36,103 @@ def show_results(
...
+    }
+    for_text = ["loss_classifier", "loss_box_reg", "loss_total"]
+
+    with torch.no_grad():
+        for (x, y), sample in pbar(dl):
+            loss = model(x, y)
+            loss = {k: float(v.numpy()) for k, v in loss.items()}
+            loss["loss_total"] = sum(loss.values())
+
+            text = ""
```
lgvaz, line 83: Is it possible somehow to move the text logic to be inside the plot function? Or maybe there is an advantage to keep it here?

icevision/models/torchvision_models/faster_rcnn/show_results.py
```diff
@@ -31,3 +36,103 @@ def show_results(
...
+    -------
+    sorted_samples: those are the samples (dictionaries) contained in the original dataset
+                    enriched with losses and sorted by `sort_by`.
+    sorted_preds: `model` predictions on `sorted_samples`. Same order as `sorted_samples`.
+    losses_stats: dictionary containing losses stats (min, max, mean, 1-25-50-75-99 quantiles)
+                  for each one of the losses.
+    """
+    model_inf = deepcopy(
+        model
+    )  # we need to create a copy of the model as `get_preds` puts it in eval mode and we need it in training mode
```
lgvaz, line 127: I have not understood yet why we need to deepcopy the model. So, `get_losses` is first called with the model in `training` mode, then `get_preds` get's called and only then the model is put into `eval` mode.

If it's necessary we can put the model in training mode again after `get_preds`

icevision/models/torchvision_models/faster_rcnn/show_results.py
```diff
@@ -31,3 +36,103 @@ def show_results(
...
+    _, preds = get_preds(model_inf, dataset)
+    sorted_samples, sorted_preds, annotations = sort_losses(samples, preds, by=sort_by)
+    assert len(sorted_samples) == len(samples) == len(preds) == len(sorted_preds)
+    show_preds(
+        samples=sorted_samples[:n_samples],
+        preds=sorted_preds[:n_samples],
+        annotations=annotations[:n_samples],
+    )
+
+    return sorted_samples, sorted_preds, losses_stats
```
lgvaz, line 138: Can we break this function into two? One that calculates these statistics and another one to plot? (so the user might want to get the statistics without plotting)

lgvaz (commented):
sup sup Francesco, I added some comments =)

This is look very nice already!

icevision/models/torchvision_models/faster_rcnn/prediction.py
```diff
@@ -70,3 +78,12 @@ def convert_raw_prediction(raw_pred: dict, detection_threshold: float):
         "bboxes": bboxes,
         "above_threshold": above_threshold,
     }
+
+
+def get_preds(model: nn.Module, dataset: Dataset, batch_size: int = 16) -> List[dict]:
```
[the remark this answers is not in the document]
FraPochetti: I get your point.
I added it just to avoid writing separately the 2 lines which the function contains.
We can keep it and make it private or I can just simply type 
```
dl = infer_dl(dataset, batch_size=batch_size)
samples, preds = predict_dl(model=model, infer_dl=dl)
```
Which one do you prefer?

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Given I am running the forward image by image, so `batch_size=1` (approach which cannot be changed) I thought there would be no benefit in moving the samples to GPU.
Just overheard for nothing gained. No?

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: > If we want all samples to have the same dropout values and all, should we move the manual seed to inside the for loop?

I checked and in-outside the loop does not change the result. The seed is set properly.

> what happens if we do:

I can try doing that indeed!

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Need to look into this.

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: `get_preds` put the model in `eval` mode, meaning that the forward will now calculate predictions and not losses.
So when calling `plot_top_losses` again, in the same notebook/script, without re-instantiating the model, the function will error out.

This is why I am copying the model.
I can indeed set `model.training = True` after `get_preds` and it should work without copying!

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Actually it is already broken in 2, as I invoke `get_losses` inside `top_plot_losses`.
I am just returning `losses_stats` as a courtesy.
But this can be avoided, e.g. we simply don't return `losses_stats` and the user will have to manually invoke `get_losses` to get stats.
It seemed more work for nothing.
Wdyt?

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: This is interesting, to say the least! Not sure what is going on...

image

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Given the screenshot I shared above, setting `model.training = True` after `model.eval()` might not work.
I need to take a closer look but we might be screwing up the internal inner workings of torchvision.

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: So this works:
```
model.eval()
model.train()
model(x, y)
```
e.g. it does not return `{}` but proper losses, but it does not solve the problem of the seed.
We can totally use this approach to avoid copying the model (see 2 comments below) but the seed will still be needed.

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Done. Removed `deepcopy` and added `model.train()`.

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Done, moved the `seed` to inside the loop.

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Also, `text` (e.g. annotations) is added to the sample dict, to facilitate retrieving the info later on.

icevision/models/torchvision_models/faster_rcnn/prediction.py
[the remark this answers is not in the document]
lgvaz: Is this function only called in one place? If so I think it's better to just have the two lines

icevision/models/torchvision_models/faster_rcnn/show_results.py
lgvaz: I think even with `bs=1` it might still be faster on GPU, we can try and see =)

icevision/models/torchvision_models/faster_rcnn/prediction.py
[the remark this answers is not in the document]
FraPochetti: Ok will do that then. Yes, it is called in just one place.

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: I will then arrange this and move the tensors to whatever device the model is on.

icevision/models/torchvision_models/faster_rcnn/prediction.py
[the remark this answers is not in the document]
FraPochetti: Done.

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Done. If the model is on GPU then the tensors will be moved to CUDA too.

icevision/models/torchvision_models/faster_rcnn/show_results.py
lgvaz: Hm... Makes sense.

What I was originally thinking is to have `plot_top_losses` receive as input the result of `get_losses`, but I think what you made makes more sense! Give me some time to think more about it :sweat_smile:

icevision/models/torchvision_models/faster_rcnn/show_results.py
lgvaz: I see your point, the only thing I have in my head is that `text` is very specific to the plotting functionality, while reading the code it's not so intuitive why this field is being built. When we get to `plot_top_losses` is also not so intuitive to find where the `annotations` are being created from. 

There is nothing wrong in the way you implemented, my only opinion is that having this snippet inside `plot_top_losses` is more intuitive and more immediate to understand what is happening.

FraPochetti: I understand. Let me see what I can do about this.

FraPochetti: I split the `text` logic into a separate function and added all the losses to the plot annotations for completeness, given we are providing `total` and/or `weighted` losses and it might be confusing what those contain.
This is the new viz.
image

FraPochetti: Let me know your thoughts.

icevision/models/torchvision_models/faster_rcnn/show_results.py
```diff
@@ -31,3 +37,120 @@ def show_results(
...
+    sorted_samples: those are the samples (dictionaries) contained in the original dataset
+                    enriched with losses and sorted by `sort_by`.
+    sorted_preds: `model` predictions on `sorted_samples`. Same order as `sorted_samples`.
+    losses_stats: dictionary containing losses stats (min, max, mean, 1-25-50-75-99 quantiles)
+                  for each one of the losses.
+    """
+    samples, losses_stats = get_losses(model, dataset)
+    samples = add_annotations(samples)
+
+    dl = infer_dl(dataset, batch_size=16)  # setting `batch_size` to 16 is arbitrary.
```
lgvaz, line 137: Should we have `batch_size` as a function parameter? The only problem that I see is that `16` might be too big of a batch size for some high resolution datasets and memory might overflow

lgvaz (commented):
This is looking very very nice!!!

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: Yes indeed. I had forgotten to do that actually :D

FraPochetti: Done

icevision/models/torchvision_models/faster_rcnn/show_results.py
lgvaz: Looks really nice!!! Can you post this image on discord so we see what other people think?

FraPochetti: Done

icevision/models/torchvision_models/faster_rcnn/show_results.py
FraPochetti: @lgvaz any thoughts on this one?

lgvaz: I think we leave it like this =)

FraPochetti:
I took the liberty of adding to the `icevision` notebooks the `OCHuman_Keypoint_Detection` one, the same already stored in `icedata`. 
I was looking for a quick starting point to add `plot_top_losses` to `keypoints_rcnn` and found nothing :) .
Kind of weird, so I added it.

icevision/models/interpretation.py
```diff
@@ -0,0 +1,244 @@
+__all__ = [
+    "sort_losses",
+    "get_stats",
+    "compute_weighted_sum",
```
ai-fast-track, line 4: What do you think about replacing `compute_weighted_sum`  by `get_weighted_sum`

icevision/models/interpretation.py
```diff
@@ -0,0 +1,244 @@
...
+    "add_annotations",
+    "extract_losses_from_samples_dicts",
```
ai-fast-track, line 6: What do you think about replacing `extract_losses_from_samples_dicts` by `get_samples__losses`?

icevision/models/interpretation.py
FraPochetti: Sure! Will do

icevision/models/interpretation.py
FraPochetti: Yep, makes sense. Will do!

icevision/models/interpretation.py
FraPochetti: Done

icevision/models/interpretation.py
FraPochetti: Done

ai-fast-track:
Thanks a lot @FraPochetti!

PR #590, opened 2020-12-13
Added label_field to VIA parser for alternate region_attribute names

jerbly:
In VGG Image Annotator v2 you can specify any name for the `region_attributes`. One of these attributes holds the `label` that IceVision requires to associate with the bbox. The previous implementation of the parser made an assumption that this attribute would be named `label` and therefore would otherwise fail.

It's also possible to add multiple `region_attributes` and they can be complex types rather than simple strings. See the `qualities` attribute from the annotator:
image
Region 1 above appears as:
```json
"region_attributes": {
  "label": "z",
  "object": "z",
  "qualities": {
    "is_rotated": true
  }
}
```
Two changes are made in this PR.
- A `label_field` parameter is added to the VIA parser which defaults to `label` for backwards compatibility.
- A `VIAParseError` exception is raised during parsing if an attempt is made to parse a non-string or missing region_attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `aggregate_records_objects` function
 
 ### Changed
+- Added `label_field` to VIA parser to allow for alternate `region_attribute` names
+ 
 ### Deleted
 
 ## [0.5.0]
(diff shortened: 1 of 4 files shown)


lgvaz (approved):
Looks perfect to me!! I'll merge it right away! 

Thanks a lot for another PR @jerbly ❤️

PR #593, opened 2020-12-16
Add `VIAMaskParser` for parsing polygons as mask annotations.

rsomani95:
In order to do this, I had to remove `VIABBoxParser`'s ability to parse bounding boxes from both `rect` and `polygon` shape attributes and assume that `rect` annotations are for bounding boxes and `polygon` annotations for masks. If not, then the polygon annotations were added as an additional bbox and this didn't allow it to be parsed due to length mismatch betwen number of boxes and masks.

I've left some of the older code commented out rather than delete it, will be happy to clean that up after some feedback.

### DIFF
--- a/icevision/parsers/via_parser.py
+++ b/icevision/parsers/via_parser.py
@@ -11,7 +11,7 @@ def via(
     img_dir: Union[str, Path],
     class_map: ClassMap,
     label_field: str = "label",
-    mask: bool = True,
+    mask: bool = False,
 ) -> Parser:
     """
     Parser for JSON annotations from the VGG Image Annotator V2.
--- a/icevision/parsers/via_parser.py
+++ b/icevision/parsers/via_parser.py
@@ -142,7 +142,7 @@ class VIAMaskParser(VIABBoxParser, MasksMixin):
                     ]
 
                     # create empty image and fill it with the polygon
-                    # mask_img = PIL.Image.new("L", get_image_size(self.filepath(o)), 0)
+                    mask_img = PIL.Image.new("L", get_image_size(self.filepath(o)), 0)
                     PIL.ImageDraw.Draw(mask_img).polygon(xy_coords, outline=1, fill=1)
 
                     # return binary mask array from created image
--- a/icevision/parsers/via_parser.py
+++ b/icevision/parsers/via_parser.py
@@ -12,6 +12,10 @@ def via(
     class_map: ClassMap,
     label_field: str = "label",
     mask: bool = False,
+    bbox_from_rect: bool = True,
+    bbox_from_polygon: bool = True,
+    mask_from_rect: bool = False,
+    mask_from_polygon: bool = False,
 ) -> Parser:
     """
     Parser for JSON annotations from the VGG Image Annotator V2.
@@ -28,8 +32,31 @@ def via(
     # Returns
         The Parser
     """
+    if mask:
+        if bbox_from_rect and bbox_from_polygon:
+            raise ValueError(
+                f"Choose bboxes from either `rect` or `polygon` "
+                "when retrieving mask annotations"
+            )
+        # user needs to pick at least one option for fetching masks
+        # for backward compatibility, both are set to False by default
+        if not mask_from_rect and not mask_from_polygon:
+            raise ValueError(
+                f"If fetching mask annotations, set `mask_from_rect` and/"
+                "or `mask_from_polygon` to True"
+            )
+
     parser_cls = VIAMaskParser if mask else VIABBoxParser
-    return parser_cls(annotations_file, img_dir, class_map, label_field)
+    return parser_cls(
+        annotations_file,
+        img_dir,
+        class_map,
+        label_field,
+        bbox_from_rect,
+        bbox_from_polygon,
+        mask_from_rect,
+        mask_from_polygon,
+    )
 
 
 class VIAParseError(Exception):
@@ -43,11 +70,19 @@ class VIABaseParser(Parser, FilepathMixin, LabelsMixin):
         img_dir: Union[str, Path],
         cls_map: ClassMap,
         label_field: str = "label",
+        bbox_from_rect: bool = True,
+        bbox_from_polygon: bool = True,
+        mask_from_rect: bool = False,
+        mask_from_polygon: bool = False,
     ):
         self.annotations_dict = json.loads(Path(annotations_filepath).read_bytes())
         self.img_dir = Path(img_dir)
         self.cls_map = cls_map
         self.label_field = label_field
+        self.bbox_from_rect = bbox_from_rect
+        self.bbox_from_polygon = bbox_from_polygon
+        self.mask_from_rect = mask_from_rect
+        self.mask_from_polygon = mask_from_polygon
 
     def __iter__(self):
         yield from self.annotations_dict.values()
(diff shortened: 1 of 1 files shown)


rsomani95:
Qualitative output check:

Screenshot 2020-12-16 at 4 53 07 PM

icevision/parsers/via_parser.py
```diff
@@ -108,3 +115,38 @@ def bboxes(self, o) -> List[BBox]:
...
+                    xy_coords = [
+                        (x, y)
+                        for x, y in zip(
+                            shape_attr["all_points_x"], shape_attr["all_points_y"]
+                        )
+                    ]
+
+                    # create empty image and fill it with the polygon
+                    # mask_img = PIL.Image.new("L", get_image_size(self.filepath(o)), 0)
+                    PIL.ImageDraw.Draw(mask_img).polygon(xy_coords, outline=1, fill=1)
```
lgvaz, line 146: Instead of returning a image mask, can we directly return a `Polygon` class?

lgvaz (commented):
An option that we have is to keep the bbox parser able to handle both "polygon" and "rect" and overwrite the function `bboxes` to only parse `rect` on the mask parser. (I'm just throwing this out there, I'm not sure this would be the best strategy)

How common it's to have bboxes as polygon? I think this is a specific case because @jerbly originally had rotated bboxes on his dataset.

@jerbly what do you think?

icevision/parsers/via_parser.py
rsomani95: The current `Polygon` class requires `points` to be vertices, like the COCO format. We'll need to convert `x,y` coordinate pairs to vertices to do that. I'm not sure how to do this, but I can look around for solutions

rsomani95:
> An option that we have is to keep the bbox parser able to handle both "polygon" and "rect" and overwrite the function `bboxes` to only parse `rect` on the mask parser. (I'm just throwing this out there, I'm not sure this would be the best strategy)

I think it would be good to have that option. We can have the following scenarios:
1. (As you described) Using only `VIABBoxParser` -- parse from either `rect` or `polygon` (not both..?)
2. Using `VIAMaskParser` with _only_ `polygon` annotations -- derive `bboxes` and `mask`s both from the same `polygon` annotation (could save a user a lot of time when annotating)
3. Using `VIAMaskParser` with both `rect` and `polygon` annotations -- this is what's been implemented in this PR as of now

rsomani95:
@jerbly I'm unable to load the test [`via_bbox.json` file](https://github.com/airctic/icevision/blob/master/samples/via/via_bbox.json). I'm using [via-2.0.10](https://www.robots.ox.ac.uk/~vgg/software/via/via.html). On trying to load the project, I get a `Cannot import project from a corrupt file` at the bottom of the browser.

jerbly:
> An option that we have is to keep the bbox parser able to handle both "polygon" and "rect" and overwrite the function `bboxes` to only parse `rect` on the mask parser. (I'm just throwing this out there, I'm not sure this would be the best strategy)

As a user you're already selecting `mask=True` so why not just push the shape parsing down to the specific class and continue to support polys converted to bboxes. This would also allow you to convert rects to masks!

jerbly:
> @jerbly I'm unable to load the test via_bbox.json file. I'm using via-2.0.10. On trying to load the project, I get a Cannot import project from a corrupt file at the bottom of the browser.

`via_bbox.json` is an annotation export file, not a project file. Try importing the annotation file.

icevision/parsers/via_parser.py
```diff
@@ -11,6 +11,7 @@ def via(
     img_dir: Union[str, Path],
     class_map: ClassMap,
     label_field: str = "label",
+    mask: bool = True,
```
jerbly, line 14: `mask` should default to `False` for backwards compat

rsomani95: Fixed in 3932668198abadb2fd9ed21afee7c3a616f1c157

rsomani95:
Made some heavy-ish changes based on the feedback. The parser is now quite flexible, `bbox` and `mask` annotations can be derived from both `rect` and/or `polygon` annotations. The choice is left entirely to the user and defaults have been set to maintain previous behavior. 

I've also separated the logic to create `BBox` or `MaskArray` from the annotations into a separate function so we can tweak that more easily in the future if needed. Masks from `rect` annotations are yet to be implemented. 

Curious to hear what you guys think :)

jerbly:
I'm considering adding keypoint parsing for VIA too so this gets pretty messy. There comes a point where the factory function is more cumbersome than just instantiating the concrete class yourself. So I think either remove the factory function or do something like this... 

Instead of all the boolean flags in the factory function we should use two enums. One for `ParserType` which IMO should be moved up into the `parser.py` for all parsers. This can be used to fix the factory function `coco` too. (Which hasn't been extended to include keypoints @lgvaz ). A second enum `VIAShape` specifically describes the shapes. The user can then provide a set of shapes to exclude from the input.

Something like this:
```python
from enum import Enum
VIAShape = Enum('VIAShape', [('POLYGON', 'polygon'), ('RECT', 'rect'), ('POINT', 'point'),
                             ('POLYLINE', 'polyline'), ('ELLIPSE', 'ellipse'), ('CIRCLE', 'circle')])
ParserType = Enum('ParserType', 'BBOX MASK KEYPOINT')

def via(
    annotations_file: Union[str, Path],
    img_dir: Union[str, Path],
    class_map: ClassMap,
    label_field: str = "label",
    parser_type: ParserType = ParserType.BBOX,
    exclude_shapes: set = None
) -> Parser:
    if parser_type is ParserType.BBOX:
        parser_cls = VIABBoxParser
    elif parser_type is ParserType.MASK:
        parser_cls = VIAMaskParser
    else:
        raise NotImplementedError()
    return parser_cls(annotations_file, img_dir, class_map, label_field, exclude_shapes)
```
As a user, if I wanted a mask parser that ignored `rect` shape attributes in the via file I would call this:
```python
p = via('file', 'dir', class_map, parser_type = ParserType.MASK, exclude_shapes = {VIAShape.RECT})
```
Then in the parser classes you can get the union set of not-supported-shapes and excluded-shapes to determine which shapes to parse. Oh, and by using the string for the enum value in `VIAShape` you can do things like this:
```python
shape = VIAShape(shape_attr["name"])
if shape not in exclude_shapes:
...
```

lgvaz:
I'm up for removing the factory function and just instantiating the class, let's go with that?

jerbly:
> I'm up for removing the factory function and just instantiating the class, let's go with that?

Yes. I think that's much cleaner. Also it means we don't need `BBoxFrom` and `MaskFrom` since the contract is now just in the concrete class. I think it means a lot of code can be deleted! (MY favourite type of coding).

lgvaz:
> a lot of code can be deleted! (MY favourite type of coding).

Specially when it's untested code, then I just see the coverage percentages going up, really nice! hahaaah

Let's do that then. @rsomani95 what do you think?

rsomani95:
Concur, I think the `Enum` approach + instantiating from class is _much_ cleaner

lgvaz:
Hi hi, I hope you both had a great new year!

I'm returning back to this, @rsomani95 should I go and make the changes or you would like to do it?

> I think the Enum approach + instantiating from class is much cleaner

Just to be clear, I think if we're instantiating directly from the class we don't need Enums anymore

rsomani95:
Hi @lgvaz, thanks. hope you had a good one too!
I’m a bit caught up with some tasks at work so feel free to make the changes if you’d like :)

jerbly:
@lgvaz - I think there's potentially still value in the VIAShape enum to tidy up the parsing and to allow the user to optionally specify shapes to exclude. But, perhaps that's an extension from initial mask support for another PR some day?

lgvaz:
@jerbly I'm a bit confused about `exclude_shapes`, in the example you provided we have:

```python
...parser_type = ParserType.MASK, exclude_shapes = {VIAShape.RECT})
```

If I understood correctly this would not parse bboxes, but to train the segmentation models we need them. Maybe what you have in mind is that if rect gets remove we calculate the bboxes from the masks?

jerbly:
@lgvaz - no, that's not what I intended. The VIA spec allows you to mark up multiple shapes like rectangles, circles, ellipses etc. The exclude_shapes set was intended to allow you to take a dataset but only parse a subset of the shapes. This is independent of the parsing type be that bboxes, mask or keypoints. Think of shapes as the input and bbox/mask/keypoint as the output.

I can have a go at a PR for this if that makes more sense? I'm just a little short on time this week.

lgvaz:
@jerbly - Aaah okay, gotcha!!

Let's do that as a subsequent PR as you suggested, I'll work and merge this one this week and then you can open a new PR next week, what do you think?

jerbly:
@lgvaz - sure, I can if you think there's value in it?

lgvaz:
@jerbly - I don't have a definitive answer right now, let me explore VIA a little bit more this weekend while I work on this PR, then I can provide you with a better answer without BSing :sweat_smile:

lgvaz:
I'm falling behind on this, but just want to let you know that I didn't forget, I'll get back to this after mmdetection support has been tackled

rsomani95:
@lgvaz @jerbly closing this as it's quite outdated with the new record API :)

Issue #585, opened 2020-12-10
Tutorial for how to use FixedSplitter 


RibenaMapleSyrup:
## 📓 New <Tutorial/Example>
It would be useful to have a tutorial on how to use FixedSplitter when parsing dataset with predefined splits for train, test and validation. 

This tutorial would be relevant to any icevision task as it focusses on preparing our dataset. 

I'm not sure which dataset would be most appropriate for this tutorial but I would like to see FixedSplitter applied to a Pascal VOC formatted dataset where splits are derived from those specified by defined ImageSets.

---
**Don't remove**
Main issue for examples: #39

RibenaMapleSyrup:
Happy to look at this!

lgvaz:
Thanks a lot @RibenaMapleSyrup! I just assigned the issue to you

Issue #91, opened 2020-06-17
Integrate YOLO


lgvaz:
## 🚀 Feature
Adds support for training YOLO. 

This can be a good experience to understand if we only need a model hub or something like we currently have for detr (where we use all their source code).

YOLO v5 is available [here](https://github.com/lgvaz/yolov5)

oke-aditya:
We would need a model hub and core API. 
Suppose some researcher tomorrow comes up with slight modification of YOLO. We can't edit it on hub right ? Since hub is community maintained. 

But yes we can edit in Core API as it is maintained by us.

Making a new hub module for adding a small new edit (e.g. backbone change, init change or layer change) would be something inefficient. But making edit to make Core API more SOTA is better.

lgvaz:
closed in favor of #506

Issue #84, opened 2020-06-16
Wandb sweeps tutorial


lgvaz:
## 📓 New <Tutorial/Example>
**Is this a request for a tutorial or for an example?**  
Tutorial.

We need a tutorial on how to use wandb sweeps, after we understand how to properly use it we can even add a helper function in the core library if it seems fitting.

---
**Don't remove**
Main issue for examples: #39

lgvaz:
@paras-jain this also has wandb that I know interests you, this is why I'm tagging 😁, pick the task you would most like to work in

oke-aditya:
Please have look at [this](https://github.com/oke-aditya/mantisshrimp_wandb). @ai-fast-track @lgvaz @paras-jain .
Let me know what improvements are needed. It is silightly complex to set up and do the job. Also, we need to make .py files as wandb does not support jupyter very well. So how does that go in our docs ?

FraPochetti:
Not relevant for the time being.

PR #594, opened 2020-12-16
Tutorial for how to use FixedSplitter

RibenaMapleSyrup:
This tutorial shows how to split a dataset using predefined imagesets as specified in issue #585. The dataset used within the tutorial is Pascal VOC 2012. This tutorial will only work using the latest icedata (the one via PyPI won't work as specified in "AttributeError: module 'icedata' has no attribute 'voc' #84") with my proposed changes in PR "Add voc data functions #91".

### DIFF
--- a/notebooks/voc_predefined_splits.ipynb
+++ b/notebooks/voc_predefined_splits.ipynb
@@ -4,6 +4,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


lgvaz (approved):
Alright, perfect!

This tutorial will be very helpful! I'll make a subsequent commit to add this to the docs

Thanks a lot!

PR #604, opened 2021-01-11
Improving `draw_sample` 

rsomani95:
PR based on the extensive discussion on [Discord](https://discord.com/channels/735877944085446747/780951885485965352/797309685115387915).

This isn't final by any means, but more of a starting point that we can all build on.

### Key features to begin with:
- [x] Custom `.ttf` font for labelling
- [x] Plot confidence predictions
- [x] "Pretty" predictions -- capitalize label, show confidence as percentage (maybe add provision for a custom callback function here)
- [x] Draw label using ~cv2~ PIL
- [x] Provide a ColorMap for each label in `ClassMap`
- [x] Expose mask `blend` parameter

### Subjective Decisions
- [x] Uniform color for all labels' text? - **Yes**
- [x] Decide on a good default font - `SpaceGrotesk Medium`

### TODOs

- ~Refactor color-map into `ClassMap` -- implemented by @aaodiall [here]~(https://github.com/aaodiall/icevision_example/blob/main/Test_Fix_Color_Icevision.ipynb)
- [x] ~After removing `cv2` dependency, fallback to PIL default font when no custom font is provided~ Add custom font that's provided with the package. What's the best way to load this by default?
- [ ] Add tests for color conversion
- [x] ~Make sure label is plotted _inside_ the bbox~
- [ ] Lots of (redundant) type casting between `PIL.Image` and `np.ndarray` happening -- clean this up

### Getting rid of OpenCV
- [x] Replace `draw_bbox` with PIL functions
- [x] In `draw_keypoints`, replace `cv2.line` with `ImageDraw.Draw.rectangle`
- [x] In `draw_keypoints`, replace `cv2.circle` with `ImageDraw.Draw.eelipse`
- [x] In `draw_mask`, get rid of `cv2.erode`. Unclear what's the best way to do this. Maybe use `PIL.ImageOps` to add a border to the mask image (shown [here](https://stackoverflow.com/questions/1995772/applying-image-decoration-border-in-python-programmatically)) alone and then do `Image.alpha_composite` ?

cc: @lgvaz @FraPochetti


FraPochetti:
Nice work!
A quick comment. 
All these changes impact `draw_sample` but unless we call the function directly, which is very rarely the case, we also need to modify the signature of all the higher-level functions (`show_records`, `show_samples`, `draw_record`, `draw_sample`, etc) to pass the relevant arguments down to `draw_sample`.
Is this correct?

rsomani95:
> All these changes impact `draw_sample` but unless we call the function directly, which is very rarely the case, we also need to modify the signature of all the higher-level functions (`show_records`, `show_samples`, `draw_record`, `draw_sample`, etc) to pass the relevant arguments down to `draw_sample`.
> Is this correct?

That's absolutely correct. We'll need to make those changes -- perhaps after we're satisfied with the underlying functionality in `icevision.visualize.draw_data` ?

FraPochetti:
Yep sounds good to me. Just wanted to have it confirmed that I was not missing anything

lgvaz (commented):
I started added comment reviews and only then I realized they were all in your `TODO / Debatable ` list :sweat_smile: 

Yeah yeah, this is a very fine starting point indeed, I'm very excited to get rid of opencv (or at least getting closer to that)

Can you post a current image on how the visualization is looking right now?

FraPochetti:
Was about to ask the same as @lgvaz :) even though the visualization does not really change at all.

rsomani95:
I realised there's an assumption I glossed over.

No big deal, I'm able to produce customised results locally, will reproduce and share a Colab notebook for you guys to play with.

rsomani95:
For documenting changes -- we moved this conversation to Discord for lower latency communication. I've updated the first comment on top with some of the main takeaways

icevision/visualize/draw_data.py
```diff
@@ -22,49 +29,115 @@ def draw_sample(
...
         sample.get("keypoints", []),
     ):
+        # random color by default
         color = (np.random.random(3) * 0.6 + 0.4) * 255
-        color = tuple(color.astype(int).tolist())
+
+        # logic for plotting specific labels only
+        # `include_only` > `exclude_labels`
+        if not label == []:
+            label_str = class_map.get_id(label)
```
rsomani95, line 80: @lgvaz A bunch of the test failures stem from my assumption that a `class_map` is always passed. If this isn't true, I could change this to something like:
```python
label_str = class_map.get_id(label) if class_map else ""    
```

icevision/visualize/draw_data.py
```diff
@@ -77,36 +150,51 @@ def draw_label(
...
-    cv2.rectangle(img, box_pt1, box_pt2, color, cv2.FILLED)
-
-    label_pt = (box_pt1[0], box_pt1[1] - 10)
-    cv2.putText(img, caption, label_pt, font, font_scale, (240, 240, 240), 2)
-
-    return img
+        return np.array(img)
 
 
 def draw_record(
```
rsomani95, line 200: Lotta code smell here. In a future PR, maybe we could:
* Directly write the addition `record.load()` in `show_record` itself so we don't need this function call
* Move to `**kwargs`?

rsomani95: Same thing applies to `draw_pred` as well

lgvaz (approved):
Alrighttttt, lets gooooooo

lgvaz:
Thanks a lot Rahuuuul!!

Issue #613, opened 2021-01-21
Don't convert Dataset sample to dict


lgvaz:
We are currently converting records to dict in `Dataset.__getitem__`. This adds unecessary code to all record mixins (`as_dict`) and breaks the neat `Record` mixins pattern. There is no reason to not keep using the records all the way in the pipeline, we just need a way to differentiate "loaded" records from unloaded ones (maybe a ENUM of states?)

potipot:
Is this connected with how data is presented from the dataloaders? I'm facing a memory leak issue on large datasets similar to [this](https://github.com/pytorch/pytorch/issues/13246). In that thread it is advised to use pure tensors in batches.

lgvaz:
@potipot Is this also related to the problem you are describing [here](https://github.com/airctic/icevision/issues/581#issuecomment-760975415)? I didn't had time to check that yet sorry.

This change was only proposed to improve the API without the memory leak in mind, but it's a good opportunity to also take a look at that.

Can you confirm that the leaks happens only works when using multiple GPUs and works fine with one?

lgvaz:
Can you also provide how many records do you have in your use case?

potipot:
Yes it is connected with that issue. 
I experienced the leak at first when using distributed training both on pl and fastai. I figured it was connected with [this issue](https://github.com/PyTorchLightning/pytorch-lightning/issues/2352) posted on pl, where the solution was to convert data presentation: 

> My data is presented via list of tensors (bboxes and labels for detection, they're variable size) and I use regular list instead of tensor here, hence the problem with multiprocessing.
I fixed my problem for now with setting sharing strategy to 'file_allocator' (default), padding variable sized lists in collate_fn to return tensor instead of list of tensors and ignore 0 label when computing loss. 

It doesnt occur on single gpu training (or rather without DDP).

> Can you also provide how many records do you have in your use case?

I'm training on COCO 2017 so its 860001 train and 36781 valid records

I experienced another issue connected with how COCOMetrics are accumulated. They are stored as a list of dictionaries AFAIR and result in many GBs of memory consumption. I tried to fixed that but work is still in progress.

lgvaz:
So, tell me if I got this correctly. The problem is the data type that is returned by the dataloader, if we return mutable structures like lists and dicts this will result in memory leakage?

lgvaz:
> I tried to fix that but work is still in progress.

[...]

PR #616, opened 2021-01-23
removed images from records accumulated in validation

potipot:
I figured out the memory leak in COCOMetrics. Records were stored as whole, together with images.


lgvaz (approved):
Good catch!

This is also somehow related to #613 where we're trying to find a unified representation for a record and it's stages.

Currently we have:
Parser -> record
record -> Dataset -> sample (with is a dict representation of the record + opened images/masks)

I think it would be better to have a single abstraction for all stages, each record could then have a internal state `loaded/unloaded` or something like that.

Anyways, thanks for the PR!

Issue #571, opened 2020-12-03
Keypoint tests uses OCHuman dataset


lgvaz:
Some of the keypoints tests are done using the OCHuman dataset, but the parser for this dataset is not implemented in `icevision` (currently it's being defined inside `conftest.py`).

Instead we should do all tests using COCO keypoints, OCHuman should be moved to `icedata`.

FraPochetti:
We need to create a keypoints-COCO-sample for that to happen, no?

lgvaz:
Correcto!

FraPochetti:
Not relevant anymore.

PR #628, opened 2021-01-28
adding class_map tests in test_parser

FraPochetti:


tests/conftest.py
```diff
@@ -81,6 +81,87 @@ def fridge_ds(samples_source, fridge_class_map) -> Tuple[Dataset, Dataset]:
     return train_ds, valid_ds
 
 
+@pytest.fixture(scope="module")
+def pennfudan_ds(samples_source) -> Tuple[Dataset]:
```
lgvaz, line 85: Let me already comment here before you continue, testing custom parsers specific to a dataset should only be done in `icedata`. Icevision should not have any idea about pennfudan.

On icevision we have parsers for common formats like VOC that are not tied to a dataset. Basically what I'm trying to say is that if you're having to create a custom parser on testing it should not be on icevision. This is also related to #571  where something similar is happening

FraPochetti: Understood. Shall I just create a couple of dummy records then and remove pennfudan?
I thought it would be easier to run tests on some real-life samples.

FraPochetti: @lgvaz :)

lgvaz: sorry for the delay! completely missed it first time, hectic day this has been

so so, you can use the samples we have for coco or voc =)

FraPochetti: Thanks!

FraPochetti: @lgvaz how about the latest commit?
Is it more along the expected lines?

lgvaz: Perfect!!!

Another option for masks, if you prefer to use real-life samples, is to use the segmentation masks we have for voc or coco. The same way we create `fridge_ds` we can create `voc_mask_ds` or something like that

FraPochetti: I think I like the hand-crafted toy `records`. They make debugging really easy and they are very intuitive.

lgvaz (changes requested):
This is exactly what I had in mind and more! Thanks a lot once more Francesco :heart: 

Some small changes before we merge:
- I think we don't need the pennfundan samples anymore right?

- can we get `_base_` inside `samples/mmdet` somehow? Maybe we can follow a similar structure with what they have in their configs - `samples/mmdet/_base_`, `samples/mmdet/faster_rcnn/faster_rcnn_r50...py` and so on

FraPochetti:
> This is exactly what I had in mind and more! Thanks a lot once more Francesco ❤️
> 
> Some small changes before we merge:
> 
> * I think we don't need the pennfundan samples anymore right?
> * can we get `_base_` inside `samples/mmdet` somehow? Maybe we can follow a similar structure with what they have in their configs - `samples/mmdet/_base_`, `samples/mmdet/faster_rcnn/faster_rcnn_r50...py` and so on

Ah sure! Let me look into this!

lgvaz (approved):
Nice! Lets go!

PR #630, opened 2021-01-30
Refactor effdet dataloaders wrapper and fix the image sizes passed to effdet

potipot:
I started working on a refactor of effdet dataloaders. I noticed many common parts in the `build_train_batch`, `build_valid_batch` and `build_infer_batch` and came up with an idea to unify this API. There is a main loop over all records in order to extract certain information about the image tensor: image itself, bboxes and labels, image sizes and scales. Whether we want to extract certain parameters depends on the train/valid/infer part. 

Main change is the `process_record` function that will be used in single iteration over all records (instead of doing it twice now, second time if we need to extract additional info not handled by the base `build_train_batch` function).

I also removed the tensor casting from the `process_record` function in order to improve code readability. The idea is to handle to_tensor casting by `collate_records`.

This is a draft PR. Looking forward for your thoughts and suggestions!


lgvaz:
I'll take a look at this today =)

icevision/models/ross/efficientdet/dataloaders.py
```diff
@@ -92,26 +92,22 @@ def build_train_batch(records, batch_tfms=None):
...
-            targets["bbox"].append(tensor([[0, 0, 0, 0]], dtype=torch.float))
-        else:
-            labels = tensor(record["labels"], dtype=torch.float)
-            targets["cls"].append(labels)
-
-            bboxes = tensor([bbox.yxyx for bbox in record["bboxes"]], dtype=torch.float)
-            targets["bbox"].append(bboxes)
+    # convert to tensors
+    batch_images = torch.stack(batch_images)
+    batch_bboxes = [Tensor(bboxes) for bboxes in batch_bboxes]
```
lgvaz, line 101: Small thing, can we use `tensor` instead of `Tensor`? Is what they suggest in the documentaiton [here](https://pytorch.org/docs/stable/tensors.html#torch.Tensor)

icevision/models/ross/efficientdet/dataloaders.py
```diff
@@ -92,26 +92,22 @@ def build_train_batch(records, batch_tfms=None):
...
+    batch_classes = [Tensor(classes) for classes in batch_classes]
 
-    images = torch.stack(images)
+    # convert to EffDet interface
+    targets = dict(
+        bbox=batch_bboxes,
+        cls=batch_classes,
```
lgvaz, line 107:
```suggestion
        cls=batch_classes
```

This way it stays in one line =)

icevision/models/ross/efficientdet/dataloaders.py
```diff
@@ -134,19 +130,19 @@ def build_valid_batch(records, batch_tfms=None):
...
-    img_sizes = [(r["height"], r["width"]) for r in records]
-    targets["img_size"] = tensor(img_sizes, dtype=torch.float)
+    (batch_images, targets), records = build_train_batch(records, batch_tfms)
 
-    targets["img_scale"] = tensor([1] * len(records), dtype=torch.float)
+    # convert to EffDet interface, when not training, dummy size and scale is required
+    targets = dict(
+        img_size=None,
+        img_scale=None,
+        **targets,
```
lgvaz, line 139:
```suggestion
        **targets
```
same here I believe

icevision/models/ross/efficientdet/dataloaders.py
```diff
@@ -163,16 +159,41 @@ def build_infer_batch(dataset, batch_tfms=None):
...
+
+    # convert to tensors
+    batch_images = torch.stack(batch_images)
+    batch_sizes = Tensor(batch_sizes)
+    batch_scales = Tensor(batch_scales)
+
+    # convert to EffDet interface
+    targets = dict(
+        img_size=batch_sizes,
+        img_scale=batch_scales,
```
lgvaz, line 175:
```suggestion
        img_scale=batch_scales
```

icevision/models/ross/efficientdet/dataloaders.py
```diff
@@ -163,16 +159,41 @@ def build_infer_batch(dataset, batch_tfms=None):
...
+
+    return (batch_images, targets), records
+
+
+def process_train_record(record) -> tuple:
+    """Extracts information from record and prepares a format required by the EffDet training"""
+    image = im2tensor(record["img"])
+    classes = (
+        record["labels"] if record["labels"] else [0]
+    )  # background and dummy if no label in record
```
lgvaz, line 186: put comment on line on top so code doesn't get broken down into multiple lines

icevision/models/ross/efficientdet/dataloaders.py
```diff
@@ -163,16 +159,41 @@ def build_infer_batch(dataset, batch_tfms=None):
...
+    )
+
+    return (batch_images, targets), records
+
+
+def process_train_record(record) -> tuple:
+    """Extracts information from record and prepares a format required by the EffDet training"""
+    image = im2tensor(record["img"])
+    classes = (
+        record["labels"] if record["labels"] else [0]
```
lgvaz, line 185: I think we should use `len(record['labels'] > 0)` to be more strict and avoid possible bugs. 

For example, what happens if for some reason `record['labels'] = None`? It's not something we are prepared for and it would mean something wrong is happening in the pipeline, it's better for that to error out here.

lgvaz (changes requested):
I really really like these changes!!! I just suggested some very small changes.

Thanks a lot Pawel!

icevision/models/ross/efficientdet/dataloaders.py
potipot: you're right this looks cleaner but at the same time is more error prone

potipot:
So as I changed `Tensor` to `tensor` invocations, the effdet tests complain that the type is wrong, "should be tf.float32". Changing it back to what it was passes the tests. Don't know yet why this is happening

icevision/models/ross/efficientdet/dataloaders.py
```diff
@@ -92,26 +92,22 @@ def build_train_batch(records, batch_tfms=None):
...
+    # convert to tensors
+    batch_images = torch.stack(batch_images)
+    batch_bboxes = [tensor(bboxes, dtype=torch.float32) for bboxes in batch_bboxes]
+    batch_classes = [tensor(classes, dtype=torch.float32) for classes in batch_classes]
 
-    images = torch.stack(images)
+    # convert to EffDet interface
+    targets = dict(
+        bbox=batch_bboxes,
+        cls=batch_classes,
```
lgvaz, line 107: removing the commas from the last item in the dict here still need to be applied :sweat_smile:

icevision/models/ross/efficientdet/dataloaders.py
```diff
@@ -92,26 +92,22 @@ def build_train_batch(records, batch_tfms=None):
...
-            targets["bbox"].append(tensor([[0, 0, 0, 0]], dtype=torch.float))
-        else:
-            labels = tensor(record["labels"], dtype=torch.float)
-            targets["cls"].append(labels)
-
-            bboxes = tensor([bbox.yxyx for bbox in record["bboxes"]], dtype=torch.float)
-            targets["bbox"].append(bboxes)
+    # convert to tensors
+    batch_images = torch.stack(batch_images)
+    batch_bboxes = [tensor(bboxes, dtype=torch.float32) for bboxes in batch_bboxes]
```
lgvaz, line 101: It's a bit more verbose, but I like the fact that we're passing `dtype` explicitly here, specially because as you pointed out `effdet` complains if the dtype is not what it expects

lgvaz (changes requested):

icevision/models/ross/efficientdet/dataloaders.py
potipot: ah, so you wanted the commas removed, not added :D

potipot: It is strange though, cause black formatter forces me to use comma in the `__all__=[..]` and then allows missing commas in a regular list or dict creation. Why is it different for the two?

potipot:
Oh this missing space in COCOMetric test keeps popping up, its really annoying, sometimes it passes others it doesn't. Any idea how to fix this for good?

lgvaz:
I know what this is related to, the original `pycocotools` has the space. Because of mmdet we started using their fork `mmpycocotools` and this one doesn't have spaces

The tests are done using `mmpycocotools`, do you by any change have `pycocotools` locally?

lgvaz (approved):
latest change removing commas looks good!

potipot:
Yes, I do have it locally. Waiting for resolution of those errors before merge?

lgvaz:
> Yes, I do have it locally
`pycocotools` or `mmpycocotools`?

For the errors just remove the spaces

potipot:
I think we are set.

lgvaz (approved):
Perfect!! We are all set indeed!! let's merge

PR #631, opened 2021-02-02
Mixins refactor

lgvaz:
Before merge:
- [x] rename `default, detect, classif` to `common, detection, classification`
- [ ] cocometrics returning 0, it was working before
- [ ] all models should be working
- [x] remove component registry
- [x] rename attributes `labels->labels_ids` and `labels_names->labels` in `BaseLabelsRecordComponent`

Open issue about:
- Xhow to have documentation for mulitple versions of the library?
- XWrite a how to transition tutorial
- Prediction memory leak with image being stored
- transforms hack, only works on a single task now
- Xshow methods (record and prediction use components)
- preds now has the info about img_size and original_img_size: bboxes, masks, keypoints, can be rescaled
- optionally don't store ground_truth (and don't store imgs on pred), implement logic for loading record when required (for showing)
- Xif record is loaded, dataset will call load again
- Xrefactor covert to coco style
- Xrefactor plot_top_losses, specially [here](https://github.com/airctic/icevision/blob/master/icevision/models/interpretation.py#L247), samples is `Record` and will not have dict attributes anymore, ping Francesco
- Xdeprecate record `as_dict` and dict like interface
- preds always hold laoded image, can cause memory leak
- X~uses `add_labels` as default to add label by names (and have `add_labels_by_id`)~ rename `labels->labels_id` `labels_names->labels`
- Xuse full word for task `classif->classification`, `detect->detection`
- `set` to ensure unique components doesn't work on composites, because components are instances and are all unique
- use mixins instead of components in album adapter?
- Transforms collect can use `record_component` directly instead of `record.detect...`
- Xclass map 3 way mapping?
- `Dataset.from_images` and `ClassMap` interaction
- `Prediction` class has to be restructured


lgvaz:
BUG: Setting an attribute on record does not set on the component. For example, in the transforms we do `record.img = tfmed_img`. This creates a new attribute `img` on record instead of setting in the appropriate component.

lgvaz:
unload record at dataloader time

lgvaz:
`ClassMapMixin` was removed from `Parser` from a reason I don't remember now. Needs to be addressed

lgvaz:
Multi task transforms is still problematic:
```python
tasks_labels = {
    'task1': [0, 1, 1], # ['cat', 'dog', 'dog']
    'task2': [1, 0], # ['car', 'bus']
    'task3': [0, 1, 2, 0], # ['can', 'box', 'bottle', 'can']
}

# albu needs a single list of labels
# labels can be *removed* if a bbox gets cut from image
# labels cannot be simply concatenated, because there is no way to go back if a label gets removed
# for example, let's say the last dog got removed, we would have this a result:
# in:  [0, 1, 1, 1, 0, 0, 1, 2, 0]
# out: [0, 1, 1, 0, 0, 1, 2, 0]
#             ^ no way to know if the dog or the car was removed
# so we transform everything into a sequence
# in: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
# out: [0, 1, 3, 4, 5, 6, 7, 8, 9]

albu_in_labels = list(range(sum(len(v) for v in tasks_labels.values())))

# and now we have to go back to the dict of tasks
# this is the complicated part...
```

lgvaz:
To implement multi task, try a structure like this:
```python
class AlbumentationsLabelsComponent(AlbumentationsAdapterComponent):
    order = 0.1

    def reset(self):
        self.tasks_labels = {}
       
    def setup_labels(self, record_component):
        self.tasks_labels[record_component.task] = record_component.labels
        
    def before_transform(self):
        # concatneate labels into one
        self.adapter._albu_in["labels"] = list(range(len(self._original_labels)))
        # use the indexes to uncocnatenate
```

PR #639, opened 2021-02-11
Add detected bbox

potipot:
Working on confusion matrix I found that this class could be useful for improving my code readability. What do you think about it? Maybe there is some better place/name for that? Generally I think it could be useful to start and then add features like iou, etc. to match targets with predictions.


lgvaz (commented):
Interesting idea... Give me some time to think about it

Basically what we have now is `record.scores`, `record.labels`, `record.bboxes`. This would eventually change that to `record.bbox.scores`, `record.bbox.labels`, `record.bbox.bboxes

Can you also share your code for the confusion matrix and why this ended up being helpful? It will help me understand the use cases of this implementation

potipot:
closing cause is not relevand with Prediction class added in 0.7.0

PR #640, opened 2021-02-16
mmdet plot_top_losses integration

FraPochetti:


lgvaz (approved):
Everything seems good!!

Should we merge this immediately or are you still planning to add something else?

I guess tests are left but not really important right now, we can open an issue for that and do it later on.

FraPochetti:
replied on Discord

PR #643, opened 2021-02-16
docs(examples): fixes undefined variable 'size' in getting_started_exp

Addono:
The variable `size` wasn't initialized, which would cause an issue at runtime.

I took a brief look at the Jupiter notebook for this example, and there's a separate cell where the `size` and `presize` are defined in a separate cell. I opted to inline this value rather than extract it as a separate value, as this example already favored having these magic constants in-line.


lgvaz (approved):
perfect!

we got a bit overwhelmed maintaing all tutorials so thanks a lot for helping with these fixes

btw, we're currently thinking of how to restructure all tutorials, if you're interested in participating be sure to join our discord server =)

PR #660, opened 2021-03-05
override find_bs with len in accumulate

### LINKED ISSUE #638: fastai efficientdet fails on learn.validate() with AttributeError: 'NoneType' object has no attribute 'shape'
## 🐛 Bug
when trying to simply validate metrics for an efficientdet model with fastai 
```python
KeyError: 'image_id'
```
```python
AttributeError: 'NoneType' object has no attribute 'shape'
```

it fails when trying to read the batch size automatically: in `accumulate, find_bs`
```python
class AvgLoss(Metric):
    "Average the losses taking into account potential different batch sizes"
    def reset(self):           self.total,self.count = 0.,0
    def accumulate(self, learn):
        bs = find_bs(learn.yb)
        self.total += learn.to_detach(learn.loss.mean())*bs
        self.count += bs
    @property
    def value(self): return self.total/self.count if self.count != 0 else None
    @property
    def name(self):  return "loss"
```
**To Reproduce**
1. Go to https://colab.research.google.com/drive/1i4aXYu4wIKA7eLUK86GwTm7lq7zku_oF?usp=sharing

potipot:
fixes #638


tests/conftest.py
```diff
@@ -70,31 +70,31 @@ def fridge_ds(samples_source, fridge_class_map) -> Tuple[Dataset, Dataset]:
         class_map=fridge_class_map,
     )
 
-    data_splitter = RandomSplitter([0.8, 0.2])
+    data_splitter = RandomSplitter((0.5, 0.5))
     train_records, valid_records = parser.parse(data_splitter)
 
     tfms_ = tfms.A.Adapter([A.Resize(IMG_SIZE, IMG_SIZE), A.Normalize()])
 
-    train_ds = Dataset(train_records[:2], tfms_)
```
lgvaz, line 78: I'm a bit hesitant of this change (removing `[:2]`) because I remember I added it so the models training check wouldn't take long, specially on the github CI that is CPU only.

But but, it seems it took 7m this time, so it should be fine

tests/conftest.py
```diff
@@ -70,31 +70,31 @@ def fridge_ds(samples_source, fridge_class_map) -> Tuple[Dataset, Dataset]:
...
-    valid_dl = efficientdet.valid_dl(valid_ds, batch_size=2)
+    train_dl = efficientdet.train_dl(train_ds, batch_size=request.param)
+    valid_dl = efficientdet.valid_dl(valid_ds, batch_size=request.param)
 
     return train_dl, valid_dl
 
 
-@pytest.fixture()
-def fridge_faster_rcnn_dls(fridge_ds) -> Tuple[DataLoader, DataLoader]:
+@pytest.fixture(params=[2, 3])
```
lgvaz, line 93: this is like `pytest.mark.parametrize`?

samples/fridge/odFridgeObjects/annotations/15.xml
```diff
@@ -0,0 +1,38 @@
+<annotation>
```
lgvaz, line 1: any specific reason why we're adding another annotation? 

the image related to this annotation seems to be an empty file

lgvaz (commented):
hey hey, the proposed changes to the callback makes sense, i just have some questions on the tests

samples/fridge/odFridgeObjects/annotations/15.xml
lgvaz: Oh, I just saw you commented on the issue:

> I will increase the size of validation set to say 3 and then test again.

tests/conftest.py
potipot: currently the fridge test sample only has 6 items so I thought it would be ok. I removed it to be able to test batch_size=3

tests/conftest.py
potipot: I think parametrize is for test functions while params for multiplying fixtures

potipot:
@lgvaz what do you think? The failing tests are not from the things I touched

lgvaz (approved):
Sorry for the delay! Just checked this and it looks good!

I also locally checkout the PR and the added image `15.jpg` is there, I don't know why github UI is saying 0 bytes

Thanks a loot! Merging it rn!

PR #677, opened 2021-03-11
via parser for 0.7

jerbly:
Updated VIA parser for the new 0.7 API.


lgvaz (approved):
looks goooood! thank you!

PR #682, opened 2021-03-14
fixes plot_top_losses with new API

### LINKED ISSUE #674: Fix plot_top_losses
`plot_top_losses` got broken because of the new refactor, the type of the data passed to it now changed, we just need to make the tests pass.

What were `samples` before is not just `Record`s and don't have a `dict` like interface anymore.

@FraPochetti would you be interested in this one?

FraPochetti:
fixes #674

### DIFF
--- a/icevision/models/interpretation.py
+++ b/icevision/models/interpretation.py
@@ -122,9 +122,6 @@ class LossesRecordComponent(RecordComponent):
     def set_losses(self, losses: Dict):
         self.losses = losses
 
-    def as_dict(self) -> dict:
-        return {"losses": self.losses}
-
 
 class Interpretation:
     def __init__(self, losses_dict, valid_dl, infer_dl, predict_dl):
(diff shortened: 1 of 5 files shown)


icevision/models/interpretation.py
```diff
@@ -113,6 +114,15 @@ def _prepend_str(d: dict, s: str):
     return {(s + "_" + k if s not in k else k): v for k, v in d.items()}
 
 
+class LossesRecordComponent(RecordComponent):
```
lgvaz, line 117: Maybe it's a better place to add this to `record_components.py`?

icevision/models/interpretation.py
```diff
@@ -113,6 +114,15 @@ def _prepend_str(d: dict, s: str):
...
+    def __init__(self, task):
```
lgvaz, line 118:
```suggestion
    def __init__(self, task=tasks.common):
```

lgvaz (changes requested):
Thanks a lot Francesco!! I added some very small suggestions

lgvaz (approved):
Looks good!! Merging!

PR #689, opened 2021-03-15
Add Fastai param_groups to all MMDetection models

ai-fast-track:
Add Fastai param_groups to all MMDetection models  enable discriminative learning in fastai library.

### DIFF
--- a/icevision/models/mmdet/utils.py
+++ b/icevision/models/mmdet/utils.py
@@ -1,6 +1,4 @@
-__all__ = [
-    "param_groups",
-]
+__all__ = ["param_groups"]
 
 from icevision.imports import *
 from icevision.utils import *
@@ -21,7 +19,7 @@ def param_groups(model):
         layers += [nn.Sequential(model.rpn_head, model.roi_head)]
     else:
         raise RuntimeError(
-            "{model} must inheret either from SingleStageDetector or TwoStageDetector class"
+            "{model} must inherit either from SingleStageDetector or TwoStageDetector class"
         )
 
     _param_groups = [list(layer.parameters()) for layer in layers]
--- a/icevision/models/mmdet/utils.py
+++ b/icevision/models/mmdet/utils.py
@@ -1,0 +1,29 @@
+__all__ = [
+    "param_groups",
+]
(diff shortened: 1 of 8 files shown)


setup.cfg
```diff
@@ -25,9 +25,9 @@ zip_safe = False
 include_package_data = True
 packages = find:
 install_requires = 
-    torch >=1.7.0,<1.8
-    torchvision >=0.8.0,<0.9
-    torchtext <0.9 # quickfixes 662
+    torch ==1.7.0
+    torchvision ==0.8.1
+    torchtext ==0.8.0 # quickfixes 662
```
lgvaz, line 30: let's not pin `torchtext` version, we only need it to be smaller than `0.9`, if the user has `0.7` there is no need to upgrade it

icevision/models/mmdet/utils.py
```diff
@@ -0,0 +1,29 @@
+__all__ = [
+    "param_groups",
```
lgvaz, line 2: remove the comma and save with black so this will stay in a single line

icevision/models/mmdet/utils.py
```diff
@@ -0,0 +1,29 @@
...
+    layers += [getattr(body, l) for l in body.res_layers]
+    layers += [model.neck]
+
+    if isinstance(model, SingleStageDetector):
+        layers += [model.bbox_head]
+    elif isinstance(model, TwoStageDetector):
+        layers += [nn.Sequential(model.rpn_head, model.roi_head)]
+    else:
+        raise RuntimeError(
+            "{model} must inheret either from SingleStageDetector or TwoStageDetector class"
```
lgvaz, line 24:
```suggestion
            "{model} must inherit either from SingleStageDetector or TwoStageDetector class"
```

lgvaz (changes requested):
hey, small suggestions =)

setup.cfg
ai-fast-track: torchtext 0.8.1 is not compatible with torch 1.7.0 whereas torchtext 0.8.0 is

Issue #702, opened 2021-03-17
Minimal BaseRecord __repr__ for a loaded image


rsomani95:
## 🚀 Feature
**Is your feature request related to a problem? Please describe.**
When an image is loaded, the __repr__ printed out on inspecting the object prints out the array representation of the image like so:
```python
BaseRecord

common: 
	- Image ID: 0
	- Filepath: /Users/rahulsomani/git/icevision-orig/samples/images/000000343934.jpg
	- Image: [[[159 158 166]
  [157 156 162]
  [162 161 167]
  ...
  [181 165 105]
  [ 99  93  61]
  [ 94  91  98]]

 [[158 157 163]
  [159 158 164]
  [157 156 162]
  ...
  [191 170 115]
  [182 166 107]
  [164 159 119]]

 [[151 150 156]
  [155 154 160]
  [158 157 163]
  ...
  [183 161 104]
  [190 163  96]
  [181 171 135]]

 ...

 [[140 113 104]
  [163 133 125]
  [158 127 122]
  ...
  [218 228 253]
  [211 225 238]
  [202 219 235]]

 [[136 114 101]
  [154 130 120]
  [149 120 112]
  ...
  [211 232 235]
  [205 228 246]
  [212 231 248]]

 [[137 109  95]
  [147 114 107]
  [149 120 114]
  ...
  [213 224 244]
  [215 229 242]
  [207 225 235]]]
	- Image size ImgSize(width=640, height=480)
detection: 
	- Labels: [4]
	- Is Crowds: [0]
	- Areas: [43522.80595]
	- BBoxes: [<BBox (xmin:175.14, ymin:175.68, xmax:496.21999999999997, ymax:415.68)>]
	- Masks: <icevision.core.mask.MaskArray object at 0x7fbb564429d0>
```

**Describe the solution you'd like**
Would be nice if it could be something like this instead:
```python
BaseRecord

common: 
	- Image ID: 0
	- Filepath: /Users/rahulsomani/git/icevision-orig/samples/images/000000343934.jpg
	- Image: 640x480 <np.ndarray> Image
	- Image size ImgSize(width=640, height=480)
detection: 
	- Labels: [4]
	- Is Crowds: [0]
	- Areas: [43522.80595]
	- BBoxes: [<BBox (xmin:175.14, ymin:175.68, xmax:496.21999999999997, ymax:415.68)>]
	- Masks: <icevision.core.mask.MaskArray object at 0x7fbb564429d0>
```

PR #705, opened 2021-03-17
ImageRecordComponent pretty `_repr`

rsomani95:
Implements #702

### DIFF
--- a/icevision/core/record_components.py
+++ b/icevision/core/record_components.py
@@ -116,8 +116,11 @@ class ImageRecordComponent(RecordComponent):
         self.composite.set_img_size(ImgSize(width=width, height=height), original=True)
 
     def _repr(self) -> List[str]:
-        height, width, dims = self.img.shape
-        return [f"Image: {width}x{height}x{dims} <np.ndarray> Image"]
+        if self.img is not None:
+            height, width, dims = self.img.shape
+            return [f"Image: {width}x{height}x{dims} <np.ndarray> Image"]
+        else:
+            return [f"Image: {self.img}"]
 
     def as_dict(self) -> dict:
         return {"img": self.img}
--- a/icevision/core/record_components.py
+++ b/icevision/core/record_components.py
@@ -117,11 +117,11 @@ class ImageRecordComponent(RecordComponent):
 
     def _repr(self) -> List[str]:
         if self.img is not None:
-            ndims = len(img.shape)
+            ndims = len(self.img.shape)
             if ndims == 3:  # RGB, RGBA
-                height, width, channels = img.shape
+                height, width, channels = self.img.shape
             elif ndims == 2:  # Grayscale
-                height, width, channels = [*img.shape, 1]
+                height, width, channels = [*self.img.shape, 1]
             else:
                 raise ValueError(
                     f"Expected image to have 2 or 3 dimensions, got {ndims} instead"
--- a/icevision/core/record_components.py
+++ b/icevision/core/record_components.py
@@ -117,8 +117,16 @@ class ImageRecordComponent(RecordComponent):
 
     def _repr(self) -> List[str]:
         if self.img is not None:
-            height, width, dims = self.img.shape
-            return [f"Image: {width}x{height}x{dims} <np.ndarray> Image"]
+            ndims = len(img.shape)
+            if ndims == 3:  # RGB, RGBA
+                height, width, channels = img.shape
+            elif ndims == 2:  # Grayscale
+                height, width, channels = [*img.shape, 1]
+            else:
+                raise ValueError(
+                    f"Expected image to have 2 or 3 dimensions, got {ndims} instead"
+                )
+            return [f"Image: {width}x{height}x{channels} <np.ndarray> Image"]
         else:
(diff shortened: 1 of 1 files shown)


icevision/core/record_components.py
```diff
@@ -116,7 +116,11 @@ def set_img(self, img: np.ndarray):
         self.composite.set_img_size(ImgSize(width=width, height=height), original=True)
 
     def _repr(self) -> List[str]:
-        return [f"Image: {self.img}"]
+        if self.img is not None:
+            height, width, dims = self.img.shape
+            return [f"Image: {width}x{height}x{dims} <np.ndarray> Image"]
```
lgvaz, line 121: Hey, tricky question: does this work with grayscale images? (I'm not sure they will have the last dimension)

rsomani95: Oh, good one. It won't work there. So, shall I do another check in here like so:

```python
if len(self.img.shape)==3:
    ...
elif len(self.img.shape)==2:
    ...
```

Can't think of another method, because there's no way of knowing the type of image otherwise. 


**RELATED:**
In the previous method (above the one highlighted) - `set_img`, there's also a line that assumes 3 dims:
```python
    def set_img(self, img: np.ndarray):
        self.img = img
        height, width, _ = self.img.shape
```

lgvaz: > In the previous method (above the one highlighted) - set_img, there's also a line that assumes 3 dims:
Hahah that would break for grayscale then as well

I have a suggestion to use instead of the if statements, but I don't know if this is actually better than use the if statements:
```python
rgb_img = np.ones((16, 16, 3))
gray_img = np.ones((16,16))

img = gray_img
height, width, channels = img.reshape(*img.shape[:2], -1).shape
```

rsomani95: Interesting.
I checked - if statements are faster (and to me, personally, clearer too)

Screenshot 2021-03-18 at 12 11 39 PM

lgvaz: What an amazing theme, it seems like code from https://carbon.now.sh/ hahahahha

Let's go with the if statements then. Just add something like:
```python
raise ValueError("Expected image to have 3 or 2 dimensions but got {len(img.shape)}")
```

Feel free to change the error message to something clearer.

rsomani95: Thanks. You can get the theme here [here](https://github.com/rsomani95/jupyter-custom-theme/) btw

Got it. Update coming right up

rsomani95: Pushed in https://github.com/airctic/icevision/pull/705/commits/2a7f262d63774ef8bf975f9c84d44d74b153edca

rsomani95:
Based on the discussion above, I suppose it makes sense to refactor a bit and check for image dimensions when doing `set_img` or `_autofix` as well? 

Currently:
```python
class ImageRecordComponent(RecordComponent):
    def __init__(self, task=tasks.common):
        super().__init__(task=task)
        self.img = None

    def set_img(self, img: np.ndarray):
        self.img = img
        height, width, _ = self.img.shape
        # this should set on SizeRecordComponent
        self.composite.set_img_size(ImgSize(width=width, height=height), original=True)

    def _repr(self) -> List[str]:
        if self.img is not None:
            ndims = len(img.shape)
            if ndims == 3:  # RGB, RGBA
                height, width, channels = img.shape
            elif ndims == 2:  # Grayscale
                height, width, channels = [*img.shape, 1]
            else:
                raise ValueError(
                    f"Expected image to have 2 or 3 dimensions, got {ndims} instead"
                )
            return [f"Image: {width}x{height}x{channels} <np.ndarray> Image"]
        else:
            return [f"Image: {self.img}"]
```

Proposal:
```python
class ImageRecordComponent(RecordComponent):
    def __init__(self, task=tasks.common):
        super().__init__(task=task)
        self.img = None

    def set_img(self, img: np.ndarray):
        self.img = img
        # Probably best not to do this every time we're setting the image
        # and check only if autofixing?
        # height, width, _ = self.get_img_dims(self.img)
        height, width, _ = self.img.shape
        # this should set on SizeRecordComponent
        self.composite.set_img_size(ImgSize(width=width, height=height), original=True)

    @staticmethod
    def get_img_dims(img: np.ndarray):
        ndims = len(img.shape)
        if ndims == 3:  # RGB, RGBA
            height, width, channels = img.shape
        elif ndims == 2:  # Grayscale
            height, width, channels = [*img.shape, 1]
        else:
            raise ValueError(
                f"Expected image to have 2 or 3 dimensions, got {ndims} instead"
            )
        return height, width, channels

    def _autofix(self):
        try:
            self.get_img_dims(self.img)
        except:
            raise AutofixAbort(f"<...>")

    def _repr(self) -> List[str]:
        if self.img is not None:
            height, width, channels = self.get_img_dims(self.)
            return [f"Image: {width}x{height}x{channels} <np.ndarray> Image"]
        else:
            return [f"Image: {self.img}"]
```

lgvaz (approved):
looks gooood

lgvaz:
thank you!!

PR #708, opened 2021-03-17
Unified API for passing backbones to MMDetection models

### LINKED ISSUE #707: Use a unified API to pass backbones to MMDetection models
Use same API interface to pass backbones in MMDetection as the one used for both torchvision and efficientdet models

ai-fast-track:
Use same API interface to pass backbones in MMDetection as the one used for both torchvision and efficientdet models

closes #707


icevision/models/mmdet/models/retinanet/backbones.py
```diff
@@ -0,0 +1,72 @@
...
+    "r101_caffe_fpn_1x",
+    "r101_fpn_1x",
+    "r101_fpn_2x",
+    "x101_32x4d_fpn_1x",
+    "x101_32x4d_fpn_2x",
+    "x101_64_4d_fpn_1x",
+    "x101_64x4d_fpn_2x",
+]
+
+r50_caffe_fpn_1x = {
```
lgvaz, line 14: Sorry for not pointing this out earlier, but I think a better idea would be to use a `dataclass` instead of a dict:
```python
@dataclass
class MMDetBackboneConfig:
    model_name: str
    cfg_filepath: str
    weights_url: str
```

icevision/models/mmdet/common/bbox/single_stage/model.py
```diff
@@ -5,26 +5,49 @@
...
+    pretrained: bool = True,
+    checkpoints_path: Optional[Union[str, Path]] = None,
+    force_download=False,
 ) -> nn.Module:
 
-    # if `cfg` argument is a path (str, Path) create an Config object from the file
-    # otherwise cfg should be already an Config object
-    if isinstance(cfg, (str, Path)):
-        cfg = Config.fromfile(str(cfg))
+    save_dir = None
```
lgvaz, line 19: why do we do `save_dir = None`?

icevision/models/mmdet/common/bbox/single_stage/model.py
```diff
@@ -5,26 +5,49 @@
...
+
+    model_name = backbone["model_name"]
+    cfg_filepath = backbone["cfg_filepath"]
+    weights_url = backbone["weights_url"]
+
+    # download weights
+    if pretrained and weights_url:
+        if checkpoints_path:
+            save_dir = checkpoints_path.model_name
```
lgvaz, line 28: should this instead be `save_dir = Path(checkpoints_path) / model_name?`

icevision/models/mmdet/common/bbox/single_stage/model.py
```diff
@@ -5,26 +5,49 @@
...
+from icevision.utils.download_utils import *
 
 
 def model(
-    cfg: Union[str, Path, Config],
+    backbone,
     num_classes: int,
-    weights_path: Optional[Union[str, Path]] = None,
+    pretrained: bool = True,
+    checkpoints_path: Optional[Union[str, Path]] = None,
```
lgvaz, line 15: instead of having `None` here you can have `'checkpoints` and you avoid the `if checkpoints_path` down there

icevision/models/mmdet/common/bbox/single_stage/model.py
```diff
@@ -5,26 +5,49 @@
...
+
+        fname = Path(weights_url).name
+        weights_path = save_dir / fname
+
+        if not weights_path.exists() or force_download:
+            download_url(weights_url, weights_path)
+
+    if cfg_filepath:
+        cfg = Config.fromfile(cfg_filepath)
+    else:
```
lgvaz, line 41: I'm not sure we need `if cfg_filepath`, if this is an invalid path the function `Config.fromfile` will already thrown an error I guess

lgvaz (changes requested):

icevision/models/mmdet/common/bbox/single_stage/model.py
ai-fast-track: We are checking if the path is empty then we raise an error with a meaningful error message. I replace the message by the following one:

```
raise RuntimeError(
            "cfg_filepath is empty. Model must have a valid configuration file"
        )
```

icevision/models/mmdet/common/bbox/single_stage/model.py
ai-fast-track: No

icevision/models/mmdet/common/bbox/single_stage/model.py
ai-fast-track: Yes but removed because of the subsequent comment

icevision/models/mmdet/common/bbox/single_stage/model.py
lgvaz: What error message do we get if you pass an invalid path to `Config.fromfile`?

icevision/models/mmdet/common/bbox/single_stage/model.py
lgvaz: Sorry?

lgvaz: Sorry?

icevision/models/mmdet/common/bbox/single_stage/model.py
ai-fast-track:
```
def check_file_exist(filename, msg_tmpl='file "{}" does not exist'):
    if not osp.isfile(filename):
        raise FileNotFoundError(msg_tmpl.format(filename))

```
Basically, if the filename is empty the mmcv message will be:
**file  does not exist**
not very helpful message there. As opposed to the one I'm adding:
**cfg_filepath is empty. Model must have a valid configuration file**

lgvaz: I'm not sure the modified error message is easier to understand. It's not clear what "cfg_filepath is empty" means, do we have a file but it's empty? 

Probably I'm not understading here, but I'm guessing the problem really is when you pass a Path that does not exist, in this case the original error message makes more sense.

I also thought that by "cfg_filepath is empty" you could also be referring that it was not set on `MMDetBackboneConfig`. But you can't instantiate `MMDetBackboneConfig` without specifying a `cfg_filepath`

icevision/models/mmdet/common/utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+    if model_type == "two_stage_detector_bbox":
+        # Sparse-RCNN has a list of bbox_head whereas Faster-RCNN has only one
+        if isinstance(cfg.model.roi_head.bbox_head, list):
+            for bbox_head in cfg.model.roi_head.bbox_head:
+                bbox_head["num_classes"] = num_classes - 1
+        else:
+            cfg.model.roi_head.bbox_head.num_classes = num_classes - 1
+
+        if weights_path is not None:
```
lgvaz, line 37: why do we do `cfg.model.pretrained = None` only if `model_type` is "two_stage_detector_bbox"?

icevision/models/mmdet/common/bbox/two_stage/model.py
```diff
@@ -5,34 +5,29 @@
...
-
-    _model = build_detector(cfg.model, cfg.get("train_cfg"), cfg.get("test_cfg"))
-
-    if weights_path is not None:
-        load_checkpoint(_model, str(weights_path))
-
-    _model.param_groups = MethodType(param_groups, _model)
-
-    return _model
+    cfg, weights_path = create_model_config(
```
lgvaz, line 19: This seems to be dead code

icevision/models/mmdet/models/mask_rcnn/backbones.py
```diff
@@ -0,0 +1,75 @@
...
+    "x101_32x4d_fpn_2x",
+    "x101_64x4d_fpn_1x",
+    "x101_64x4d_fpn_2x",
+]
+
+from icevision.models.mmdet.utils import *
+
+
+r50_caffe_fpn_1x = MMDetBackboneConfig(
+    model_name="mask_rcnn",
```
lgvaz, line 18: `model_name` seems unnecessary, we only have it here because to do `save_dir = Path("checkpoints") / model_name` under `create_model_config`. But `weights_url` already have the model name included, i.e. `mask_rcnn_r50_caffe_fpn...`

Also, we can extract `model_name` from `cfg_url` if we really want to

icevision/models/mmdet/utils.py
```diff
@@ -25,3 +37,42 @@ def param_groups(model):
...
+    force_download=False,
+):
+
+    model_name = backbone.model_name
+    cfg_url = backbone.cfg_url
+    weights_url = backbone.weights_url
+
+    # download weights
+    if pretrained and weights_url:
+        save_dir = Path("checkpoints") / model_name
```
lgvaz, line 62:
```suggestion
        save_dir = Path(checkpoints_path) / model_name
```

lgvaz (changes requested):

icevision/models/mmdet/common/utils.py
ai-fast-track: That line should be common to all models. fixed now

icevision/models/mmdet/models/mask_rcnn/backbones.py
ai-fast-track: That's seems to be true for all MMDetection pretrained weights filenames. But if a user create its pretrained  pretrained weights filename without including the model name we cannot extract the name anymore.

icevision/models/mmdet/common/bbox/single_stage/model.py
ai-fast-track: `Raise Error` removed: handled by dataclass.

PR #712, opened 2021-03-19
Visualisation improvements

rsomani95:
* leverage new API for plotting all components
* dynamic label padding with full control
* thin / thick / no label border
* label border color
* make `prettify_funcs` a list to enable more functionality
* optionally include task names in labels
* expose all args only in `draw_sample`, use `**kwargs` elsewhere
* TODO: multi task classif labels

It's getting pretty gnarly to support all sorts of different use cases. I held off on implementing multi-task support for classification. I think that warrants a longer discussion as there's a lot of factors to consider.


rsomani95:
I botched this. Will redo and push another one

Issue #710, opened 2021-03-18
Improve visualization


lgvaz:
Improvements:
* Meme like border around text
* Classif labels on top-left or bottom-left
* New line for multiple labels in classif
* padding for labels from bbox edge

@rsomani95 Interested in this one? :sweat_smile:

rsomani95:
Sure. You can assign this to me

rsomani95:
Also, dynamic padding for labels from bbox edge

lgvaz:
> Also, dynamic padding for labels from bbox edge

I edit the the list and added this

lgvaz:
closed with #713

PR #713, opened 2021-03-19
Visualisation improvements

rsomani95:
Addressed #710 

[x] Border around labels
[x] Classif labels on **top-left**
[x] Padding for labels from bbox_edge

Not included:
* Classif labels in new lines for multiple tasks

This requires additional discussion. There's two levels of complexity:
1. We don't have a `MultiLabelClassificationLabelsComponent`
2. We need to track all classification labels plotted, and incrementally add the adjust the height being passed to `draw_label`
    * `draw_label` doesn't accept `x` and `y` params, so we'll need to create dummy `bbox`es
    * If it's a multi label component, do we want to put all labels from that task in one line or a separate line for each?


lgvaz:
I'm trying this on `custom_parser.ipynb` and it seems not to displaying the annotations:
image

lgvaz (approved):
coooool, working now! merging!

PR #714, opened 2021-03-19
Add MultiLabel Classification Component

rsomani95:


icevision/core/record_components.py
```diff
@@ -275,8 +276,45 @@ def setup_transform(self, tfm) -> None:
 
 
 class ClassificationLabelsRecordComponent(BaseLabelsRecordComponent):
```
rsomani95, line 278: should we rename this to `SingleLabelClassificationLabelsRecordComponent`?

icevision/core/record_components.py
```diff
@@ -275,8 +276,45 @@ def setup_transform(self, tfm) -> None:
...
+        super().__init__(task=task)
+
+
+class MultiLabelClassificationLabelsRecordComponent(BaseLabelsRecordComponent):
+    """Component for multi-label classification
+    * Doesn't allow you to `add_labels`
+    * One-hot encodes labels based on `class_map`
+    """
+
+    is_multilabel = True
```
rsomani95, line 291: `is_multilabel` is added for the ability to recognise the component during data loading time.

We could get rid of this and instead do
```python
if composite.get_component_by_type(MultiLabelClassificationLabelsRecordComponent)
```

icevision/core/record_components.py
```diff
@@ -275,8 +276,45 @@ def setup_transform(self, tfm) -> None:
...
+        self.is_one_hot_encoded = True
+
+    # def _autofix(self):
+    #     # Unnecessary?
+    #     if not self.is_one_hot_encoded:
+    #         self.to_one_hot()
+
+    def add_labels(self, labels):
+        # TODO: Should we allow this to be used and fix in `_autofix`?
+        raise NotImplementedError(f"Use `set_labels` for multi-label classification")
```
rsomani95, line 311: We could leave support for `add_labels`, but then they will not be one hot encoded.
The only scenario where this it's safe to support this is if we know for a fact that the user will use `autofix`

rsomani95 (commented):
Left a few pointers to consider

rsomani95:
Closing this. reopened in #820

PR #717, opened 2021-03-21
Use a unified API to pass backbones to EfficientDet models

### LINKED ISSUE #716: Use a unified API to pass backbones to EfficientDet models
## 🚀 Feature
Use same API interface to pass backbones in EfficientDet as the one used for both torchvision and MMDetection models

ai-fast-track:
Use same API interface to pass backbones to EfficientDet models as the one used for both torchvision and MMDetection models

closes #716


lgvaz (approved):
All good, merging right now! =)

PR #728, opened 2021-03-25
Standardize backbone argument in model method

### LINKED ISSUE #727: Standardize backbone argument in model method
## 🐛 Bug
**Describe the bug**
When creating a torchvision model, it fails because `backbone` argument is passed as function, and `model()` method expects an nn.Module. The `backbone` method need to be called before it is passed to the `model()` method

ai-fast-track:
Backbones are passed as methods. the torchvision `model()` method expect a nn.Module. Therefore, the `backbone` argument is now called when passed to the `model()` method.

closes #727


lgvaz:
resolved by #732

PR #737, opened 2021-03-29
move the _unload method

potipot:
from FilepathRecordComponent to ImageRecordComponent to avoid memory leak when the latter exists without the first


lgvaz (approved):
niceeee, thank you!

Issue #706, opened 2021-03-17
Update default parsers with `is_new`


lgvaz:
Make use of the recent implemented parameter `is_new` that is passed to `Parser.parse_fields`

Functions like `record.set_filepath`, `record.set_img_size`, now only need to be called once, refer to [this](https://github.com/airctic/icevision/blob/master/notebooks/custom_parser.ipynb) notebook for an example.

We need to update or default parsers to make use of this functionality: 
- [ ] `coco_parser.py`
- [ ] `voc_parser.py`
- [ ] `via_parser.py`

aaodiall:
I will look this on the week end.

lgvaz:
@aaodiall Thanks a lot! I just assigned the issue to you

PR #738, opened 2021-03-29
make us of is_new argument in parsers

potipot:
fixes https://github.com/airctic/icevision/issues/706


lgvaz (approved):
Looks goood!

Thanks Pawel!

PR #746, opened 2021-03-31
unloads records after collate

lgvaz:


potipot:
This should work! We should think about what happens in inference though. The same batch building logic happens for all, and this way we will always unload the records. Maybe the paths should be split for inference and training/validation here?

Issue #692, opened 2021-03-16
Predict memory leak


lgvaz:
## 🐛 Bug
The prediction contains a reference to the image array. If any stage of the workflow starts accumulating the predictions objects (as it currently happens in `COCOMetric` this will result in a memory leak.

`COCOMetric` currently uses a "quickfix" to make sure the images are not being stored (take a look at `accumulate`), but we need a more general solution.

rsomani95:
Based on our joing session the otherday on the machine with 16GB RAM, it looks like there's memory leaks happening elsewhere too, right?

As a refresher -- we'd turned off `COCOMetrics` and the process kept crashing faster proportial with `num_workers` in the `DataLoader`

lgvaz:
Yeah, we need to investigate this deeper. Let's have a call where we use a profiling tool like described [here](https://medium.com/zendesk-engineering/hunting-for-memory-leaks-in-python-applications-6824d0518774) to find the issue.

potipot:
check if is solved by https://github.com/airctic/icevision/pull/737

potipot:
Do you know if this is specific for pytorch lightning or fastai? I'm facing this issue too and investigating it RN

potipot:
I found that the problem comes from record loading. In fact, the memory use grows only in the 0th epoch and then keeps constant. This led me to thinking that the Dataset is loaded into memory as a whole and this is indeed what happens.

When calling `Dataset`'s getitem here:
```python
    def __getitem__(self, i):
        record = self.records[i].load()
        if self.tfm is not None:
            record = self.tfm(record)
        return record
```
we change the state of records, which are a field of `Dataset` class and they persist in memory as long as the Dataset and Dataloader. 
When I added a manual record unloading at the end of training loop, the memory build-up vanishes:
```python
    def training_step(self, batch, batch_idx):
        (xb, yb), records = batch
        preds = self(xb, yb)

        loss = efficientdet.loss_fn(preds, yb)

        for k, v in preds.items():
            self.log(f"train/{k}", v)
        [record.unload() for record in records]
        return loss
```

Perhaps we should have a custom Data fetcher that takes care of the cleanup at the end of training loop? Or maybe this is already implemented but operates on a deepcopy of the record rather than the object bound with `Dataset` and `Dataloader` ?

lgvaz:
Interesting question: if we use coco metric, does it solve this issue? Because coco metric does call `unload` hahaha

potipot:
[...]

PR #747, opened 2021-04-01
Add record unloading & refactor common_build_batch

potipot:
Implemented the discussed changes, mitigates the excessive leak discussed in https://github.com/airctic/icevision/issues/692

1. Unload records is a decorator of `collate_fn`, which is now put in `transform_dl`.
2. `apply_batch_tfms` (previously `common_build_batch`) is now a decorator function that is run on top of each `build_batch` function that we define within dataloaders. It is used in a single place so no repetition takes place and works in the same pattern as before. 
3. Changed plot_top_losses to reaload the now unloaded ground_truth images
3. Also fixed some smaller bugs like in icevision/models/torchvision/keypoint_rcnn/dataloaders.py@104 where the tfms were not passed at all.
4. Renamed samples or dataset to records in torchvision dataloaders for consistency.
5. Updated the docs

### DIFF
--- a/icevision/models/interpretation.py
+++ b/icevision/models/interpretation.py
@@ -128,8 +128,9 @@ class Interpretation:
         losses_dict["loss_total"] = sum(losses_dict.values())
         return losses_dict
 
-    def _loop(self, dl, model, losses_stats, device):
+    def _loop(self, dl, model, losses_stats):
         samples_plus_losses = []
+        device = model_device(model)
 
         with torch.no_grad():
             for (x, y), sample in pbar(dl):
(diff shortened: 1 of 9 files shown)


icevision/models/interpretation.py
```diff
@@ -236,14 +235,19 @@ def plot_top_losses(
             ann2 = "\n".join(ann[4:])
             anns.append((ann1, ann2))
 
+        # reload only the displayed images
+        displayed_preds = [
+            Prediction(pred=p, ground_truth=s.load())
```
lgvaz, line 240: Do we have the problem here where the predictions and even the labels in the ground truth will be related to the transformed img but `s.load()` will load the original img?

lgvaz (commented):
There might be something non-intuitive happening here: when using `predict_dl` the record will be unloaded but if you use `predict` it will not.

Does it make sense not to call `unload_records` in `infer_dl`?

icevision/models/interpretation.py
lgvaz: just confirmed that we do =x

potipot: yeah this is my suboptimal solution to this general problem

potipot:
> There might be something non-intuitive happening here: when using `predict_dl` the record will be unloaded but if you use `predict` it will not.
> 
> Does it make sense not to call `unload_records` in `infer_dl`?

Yeah we could do that, but then the entire inference dataset will persist in memory and cause a leak. Can work for single image, but then not for an entire dataset. Don't know how to find a balance here

lgvaz (approved):
looks goood!

Issue #715, opened 2021-03-20
Deployed docs do not match current release


jerbly:
## 🐛 Bug
Docs are deployed whenever we push to master. This means the content may not match what is released and lead to confusion. 

For example, right now the [custom parser tutorial live on the website](https://airctic.com/custom_parser/) contains the new API syntax for the unreleased version 0.7. This will not match what you need to do from the released version.

Perhaps we should only deploy on release?

lgvaz:
We need to get the version selection option on our docs, similar to what we have in lightning [here](https://pytorch-lightning.readthedocs.io/en/stable/).

Is this something done by the host (github/readthedocs) or by the documentation engine (mkdocs/sphinx)?

jerbly:
readthedocs does it: https://docs.readthedocs.io/en/stable/versions.html
Looks like we're using Material for MkDocs at the moment. There's a guide on versioning here: https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/ - I'll see if I can figure out what we need to do.

lgvaz:
So what you're saying is that we can do either with readthedocs OR with mkdocs?

jerbly:
Yeah... I was going to make a quick project to test how it works with mkdocs-material since that's what we're using for IceVision at the moment.... seems the quickest win. You end up with a little drop down with the version number in it top-left. See: https://jimporter.github.io/bfg9000/latest/ for example.

lgvaz:
I took a look at the plugin [here](https://github.com/jimporter/mike), so if I understood correctly, this is what we will have to do:

We have the github workflow `mk-docs-build.yml` that currently does `mkdocs gh-deploy --force` every-time we push to master, we would change that to `mike deploy 0.7 latest` (question: can we keep re-deploying to 0.7 on every push to master?).

I would actually like to do `mike deploy latest` if possible, and then on a release (which we trigger on github) we would do `mike deploy version` (the version comes from the tag of the release).

jerbly:
I've been looking at this too. And I think what you've said there is correct. I think a nice pattern is to have a `latest` and a `dev` like the https://jimporter.github.io/bfg9000/latest/ (and it's totally weird but the version numbers on that project happen to match exactly where we are!!) So exactly like bfg9000 we would have `0.6 (latest)` and `0.7 (dev)` - (note, this sets the next version ahead of time that I talked about in another thread somewhere) and it's `dev` that follows `master`.

[...]

PR #748, opened 2021-04-04
Doc site versioning, issue #715

jerbly:
This PR adds documentation site versioning using an updated mkdocs-material and [mike](https://github.com/jimporter/mike). The mike concept is to align your documentation with your releases and not to go back and edit a released version. Mike and the automation in this PR is in line with the IceVision release pattern of a single mainline based development and release process; all releases are sourced from the trunk.

Pull-requests, pushes and releases are all covered in this single action:

- `pull_request` - deploy the docs locally only in order to expose any errors
- `push` - deploy the docs as `dev` and push to the site branch
- `release` - find the latest release version, deploy the docs as that version and push to the site branch

The site branch is configurable through the env in the action. This allows for the docs to be pushed to a different branch than the live `gh-pages`. This is a contingency in case something goes wrong in the early days of this new documentation so we can easily switch which branch is published to the web.

### DIFF
--- a/.github/workflows/mk-docs-build.yml
+++ b/.github/workflows/mk-docs-build.yml
@@ -5,6 +5,11 @@ on:
     branches: [ master ]
   push:
     branches: [ master ]
+  release:
+    types: [ created ]    
+
+env:
+  SITE_BRANCH: ${{ 'gh-pages-ver' }}
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


lgvaz (approved):
Looks good! Thanks a lot Jeremy! Let's cross our fingers and see what happens!

PR #759, opened 2021-04-09
Fixed show_preds and added show_batch and in MMDetection 

ai-fast-track:
`show_preds()` load images and display them like before. This fix the issue related to unloading images in record. I also added  `show_batch()` method

### DIFF
--- a/icevision/models/mmdet/common/mask/prediction.py
+++ b/icevision/models/mmdet/common/mask/prediction.py
@@ -24,7 +24,7 @@ def _predict_batch(
     device = device or model_device(model)
     batch["img"] = [img.to(device) for img in batch["img"]]
 
-    raw_pred = model(return_loss=False, rescale=False, **batch)
+    raw_preds = model(return_loss=False, rescale=False, **batch)
     return convert_raw_predictions(
         batch=batch,
         raw_preds=raw_preds,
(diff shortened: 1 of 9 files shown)


icevision/models/mmdet/common/mask/show_batch.py
```diff
@@ -0,0 +1,18 @@
...
+from icevision.visualize import *
+
+
+def show_batch(batch_and_records, ncols: int = 1, figsize=None, **show_samples_kwargs):
+    """Show a single batch from a dataloader.
+    # Arguments
+        show_samples_kwargs: Check the parameters from `show_samples`
+    """
+    batch, records = batch_and_records
+
```
lgvaz, line 13: Do this function also needs:
```python
    if isinstance(batch["img"], list):
        batch = {
            "img": batch["img"][0],
            "img_metas": batch["img_metas"][0],
        }
```

If so, is this exactly the same as we have for box? We can maintain a single function under `common` and just export it from here

icevision/models/mmdet/common/mask/prediction.py
```diff
@@ -99,6 +142,16 @@ def convert_raw_prediction(
     pred.detection.set_masks(keep_masks)
     pred.above_threshold = keep_mask
 
+    if keep_image:
+        tensor_image = sample["img"]
+        if revert_rgb_channels:
```
lgvaz, line 147: What's the scenario where we don't want to `rever_rgb_channels`?

lgvaz (approved):

PR #760, opened 2021-04-09
add show_batch + fix plot_top_losses + fix predictions

FraPochetti:


lgvaz (approved):
Yeaaah, looks goooood!

PR #807, opened 2021-04-19
fixes draw methods for grayscale images

### LINKED ISSUE #806: draw_bbox fails for grayscale images
## 🐛 Bug
`draw_data` contains the line: `img = PIL.Image.fromarray(img)`. If the image is grayscale and in the format `(h,w,1)` (which is what icevision expects` it fails with the error:
```python
TypeError: Cannot handle this data type: (1, 1, 1), |u1
```

---
**More info:**
https://stackoverflow.com/questions/63367136/pil-typeerror-cannot-handle-this-data-type-1-1-1-u1

lgvaz:
closes #806

### DIFF
--- a/icevision/visualize/draw_data.py
+++ b/icevision/visualize/draw_data.py
@@ -368,7 +368,7 @@ def draw_bbox(
     # Calculate image dimensions
     dims = sorted(img.shape, reverse=True)
     color = as_rgb_tuple(color)
-    img = PIL.Image.fromarray(img)
+    img = PIL.Image.fromarray(img.squeeze())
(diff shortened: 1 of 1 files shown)


FraPochetti:
I think this can be easily merged.

lgvaz:
@FraPochetti is this still a thing?

FraPochetti:
> @FraPochetti is this still a thing?

We never got back to it.
Not sure it is relevant actually.

rbavery:
this seems related: https://github.com/airctic/icevision/issues/1080

PR #808, opened 2021-04-21
autofix multiprocessing

lgvaz:


lgvaz:
From AdamF on discord:

Ahh. I missed that PR! I'm not quite sure why it uses ProcessPoolExecutor (which is built on top of multiprocessing). I think that multiprocessing.Pool().map is possibly a bit simpler. 

But we should be sure that the num_workers (used in the PR) has the same semantics as in pytorch and dataloaders. In the PR, num_workers is simply passed on as the value of max_workers. In pytorch, num_workers == 0  means run in the current process, but max_workers==0 raises a value error. So, there should be a conditional num_workers==0 means plain old map, num_workers >= 0 means user Pool().map.

PR #810, opened 2021-04-21
Add confusion matrix as metric

potipot:
I finally got myself together and updated the Confusion Matrix to the newest icevision version.
Here is a colab showing how it works: https://colab.research.google.com/drive/1bEL4nlDNx6J0X6_x7O7CFvvBByHagOyb?usp=sharing

Whats missing?
- confusion matrix logging

The backbone is ready with SimpleConfusionMatrix.log() but the current icevision metric design doesn't allow to log objects other than dict or floats in case of fastai. To enable logging a refactor of the Metric is required.

I also changed the metric testing for efficientdet and think this could be reused for other used libraries. Each of them could have test_metric.py file and verify that the expected behaviour is true for all of them.


lgvaz (commented):
Hii! I like the additions here!

Before I jump into a deeper review of the code, can we make sklearn not a hard dependency? By that I mean, if you install icevision minimal (without `[all]`) I think the library will fail the import because it will try to import sklearn

We just have to change the imports around a bit to make sure `__init__` doesn't auto import `sklearn` when it's not present

We actually need a more serious discussion on how we want to handle the namespaces, we can do that on discord

potipot:
@lgvaz commits got finally updated

lgvaz:
hey hey Pawel, tests are still failing, are all the commits in this PR?

potipot:
ah yeah its due to circular imports.. I'm not sure I am handling the namespace in the right way. fix is on the way

icevision/metrics/confusion_matrix/confusion_matrix.py
```diff
@@ -0,0 +1,153 @@
+__all__ = ["SimpleConfusionMatrix"]
+
+from icevision.data.prediction import Prediction
+from icevision.metrics.metric import Metric
+from icevision.imports import *
+from icevision.metrics.confusion_matrix.confusion_matrix_utils import *
+from pytorch_lightning import loggers as pl_loggers
```
lgvaz, line 7: this import is dangerous, if lightning is not installed this will thrown an error when doing `from icevision.all import *`

icevision/metrics/confusion_matrix/confusion_matrix.py
```diff
@@ -0,0 +1,153 @@
...
+                cm = cm / cm.sum(axis=1, keepdims=True)
+            elif normalize == "pred":
+                cm = cm / cm.sum(axis=0, keepdims=True)
+            elif normalize == "all":
+                cm = cm / cm.sum()
+        cm = np.nan_to_num(cm)
+        return cm
+
+    def log(self, logger_object) -> None:
+        if isinstance(logger_object, pl_loggers.WandbLogger):
```
lgvaz, line 147: these soft imports get complicated here right? I'm also not sure rn what is the best way to handle this, but this function will error out if lightning is not installed

icevision/metrics/confusion_matrix/confusion_matrix.py
lgvaz: so, I think we can delete this line, `pl` will already come from `imports`

icevision/metrics/confusion_matrix/confusion_matrix.py
lgvaz: here we can do the following:
```python
if SoftDependencies.pytorch_lightining:
    if isinstance(logger_object, pl.loggers.WandbLogger):
        # do the stuff
```

Doesn't look good and I don't think it's a sustainable solution, but works for now?

lgvaz (changes requested):
I think we still have a problem with the way imports are done, but I'm not sure there is a quick fix that is not hacky.

icevision/metrics/confusion_matrix/confusion_matrix.py
potipot: i disabled logging at all cause it is not supported in current metric design anyway

lgvaz (approved):
lets mergeee! thanks!

PR #811, opened 2021-04-21
Update Dataset to pass the model's class map to show_preds and display the string labels in the prediction results

elmourr:
This enables passing the model's class labels to an image dataset when invoking:  dataset.from_images 
and to display the labels along with the BBoxes when running inference with show_pred.
Method:  
infer_neg_ds = Dataset.from_images([img], tfms, class_map=class_map)
show_preds(preds=preds[:6], class_map = class_map,denormalize_fn=denormalize_imagenet,ncols=3,)

### DIFF
--- a/icevision/data/dataset.py
+++ b/icevision/data/dataset.py
@@ -41,7 +41,7 @@ class Dataset:
         return f"<{self.__class__.__name__} with {len(self.records)} items>"
 
     @classmethod
-    def from_images(cls, images: Sequence[np.array], tfm: Transform = None):
+    def from_images(cls, images: Sequence[np.array], tfm: Transform = None, class_map: ClassMap = None):
         """Creates a `Dataset` from a list of images.
 
         # Arguments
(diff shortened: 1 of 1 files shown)


FraPochetti:
Thanks for the PR!

elmourr:
> Thanks for the PR!

not at all, i's been great!

lgvaz (commented):
Hii! Thanks a lot for the PR

I'm not sure we can go ahead with this (maybe as a quickfix?), the issue it's not safe to assume that our record will contain the `detection` task, you can be doing classification for example.

Hmm... Maybe because the library is mainly about the detection now this would be an okay assumption to make

FraPochetti:
> Hmm... Maybe because the library is mainly about the detection now this would be an okay assumption to make
I think it could be ok as a quickfix, yes.

lgvaz:
@elmourr hey, we can merge this one, but I believe you will need to open a new PR to merge into master like I explained in the other one

PR #812, opened 2021-04-21
Update show_data so that 'class_map' can be passed to show_preds --> see #811

elmourr:
this update allow to pass class_map to show_pred and to display the predictions with bounding boxes and string labels:
show_preds( preds=preds, class_map = class_map,denormalize_fn=denormalize_imagenet, ncols=3,)

### DIFF
--- a/icevision/visualize/show_data.py
+++ b/icevision/visualize/show_data.py
@@ -25,7 +25,7 @@ def show_sample(
     display_keypoints: bool = True,
     show: bool = True,
     ax: plt.Axes = None,
-    figsize: Tuple[int, int] = None,
+    figsize: Tuple[int, int] = None, 
     **draw_sample_kwargs,
 ) -> None:
     img = draw_sample(
--- a/icevision/visualize/show_data.py
+++ b/icevision/visualize/show_data.py
@@ -25,7 +25,7 @@ def show_sample(
     display_keypoints: bool = True,
     show: bool = True,
     ax: plt.Axes = None,
-    figsize: Tuple[int, int] = None, 
+    figsize: Tuple[int, int] = None,
     **draw_sample_kwargs,
 ) -> None:
     img = draw_sample(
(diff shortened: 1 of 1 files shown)


elmourr (commented):
please ignore this commit

icevision/visualize/show_data.py
```diff
@@ -67,6 +67,7 @@ def show_record(
 
 def show_pred(
     pred: Prediction,
+    class_map: Optional[ClassMap] = None,
```
lgvaz, line 70: Heh, so the `class_map` traces on all show functions are actually a remnant of the latest refactor - they are not required anymore, the class_map is always inferred from the record, actually what we should do is remove this parameter from the functions haha

icevision/visualize/show_data.py
```diff
@@ -83,6 +84,7 @@ def show_pred(
         display_bbox=display_bbox,
         display_mask=display_mask,
         denormalize_fn=denormalize_fn,
+        **draw_sample_kwargs,
```
lgvaz, line 87: this is a good catch yeah, once you remove the `class_map` parameter we merge this one =)

icevision/visualize/show_data.py
elmourr: ahh ok - good to know, would make sense too. good catch

icevision/visualize/show_data.py
elmourr: ok will do

elmourr (commented):
done - Removed :+1: 
class_map: Optional[ClassMap] = None,

lgvaz (commented):
Alright!! Seems good!

Now, the only thing before we merge, which I just realized, is that this PR is going to merge into the branch `parser_multiprocessing`, see it here:
image

You should change it to merge into master. I think the easiest way to do this, since it's a small change, is just to open another PR and be sure to select "into master" on the github UI

PR #813, opened 2021-04-22
Add Dev Docker Image

rsomani95:
Discussion: https://discord.com/channels/735877944085446747/780951885485965352/834668110166229053


FraPochetti:
Not sure if it makes sense to merge this one.
installation instructions are so rapidly and constantly changing that we might just put some effort into the installation scripts rather than Docker.

PR #820, opened 2021-04-24
Multi-Label Support for Classification Component + Autofix

rsomani95:
* Add `one_hot_encoded()` method for multi-label classification
* Check that we're not setting more than one label in single-classification mode
* Write some basic tests


icevision/core/record_components.py
```diff
@@ -278,8 +278,23 @@ def setup_transform(self, tfm) -> None:
 
 
 class ClassificationLabelsRecordComponent(BaseLabelsRecordComponent):
-    def __init__(self, task=tasks.classification):
+    def __init__(self, task=tasks.classification, is_multilabel: bool = False):
         super().__init__(task=task)
+        self.is_multilabel = is_multilabel
+
+    def _load(self):
+        if not self.is_multilabel and len(self.labels) > 1:
```
lgvaz, line 286: Can we move this to autofix instead? The reason being that on `_load` this will cause an error while the model is training (and interrupt training) instead of just finding all the errors at once at parsing time,

rsomani95: Addressed in f22fc60355d4463c269cf6694ea127e2e0c354e7

rsomani95:
@lgvaz I think this is ready to merge now

lgvaz:
Lets goo!!

PR #821, opened 2021-04-24
Inability to Add Invalid Values To Class Map 

### LINKED ISSUE #819: Adding Non Unique Names To ClassMap Must Be Illegal
Currently, `add_name` doesn't check if the added `name` is a duplicate or not. This should be illegal, as classes must be unique

https://github.com/airctic/icevision/blob/82f3c7322d5adbc41d6a1b7b9e4327eb32ad23c4/icevision/core/class_map.py#L47-L51

rsomani95:
Fixes #819


lgvaz (approved):
Perfecto!! Thanks Rahuul!

Issue #826, opened 2021-04-28
`draw_sample` draws multiple classification records over each other


rsomani95:
Currently, if we have multiple classification components in a record, it draws the labels all to the top left. Here's a record for example (from the `ExDark` dataset):
```python
common: 
	- Image ID: 4715
	- Filepath: /Users/rahulsomani/datasets/ExDark/images/Cup/2015_04714.JPEG
	- Image: None
	- Image size ImgSize(width=500, height=333)
location: 
	- Class Map: <ClassMap: {None: 0, 'Indoor': 1, 'Outdoor': 2}>
	- Labels: [1]
detection: 
	- Class Map: <ClassMap: {None: 0, 'Bicycle': 1, 'Boat': 2, 'Bottle': 3, 'Bus': 4, 'Car': 5, 'Cat': 6, 'Chair': 7, 'Cup': 8, 'Dog': 9, 'Motorbike': 10, 'People': 11, 'Table': 12}>
	- Labels: [8, 8, 8]
	- BBoxes: [<BBox (xmin:316, ymin:180, xmax:473, ymax:302)>, <BBox (xmin:237, ymin:20, xmax:273, ymax:70)>, <BBox (xmin:204, ymin:29, xmax:233, ymax:70)>]
lightning: 
	- Class Map: <ClassMap: {None: 0, 'Low': 1, 'Ambient': 2, 'Object': 3, 'Single': 4, 'Weak': 5, 'Strong': 6, 'Screen': 7, 'Window': 8, 'Shadow': 9, 'Twilight': 10}>
	- Labels: [2]
object_class: 
	- Class Map: <ClassMap: {None: 0, 'Bicycle': 1, 'Boat': 2, 'Bottle': 3, 'Bus': 4, 'Car': 5, 'Cat': 6, 'Chair': 7, 'Cup': 8, 'Dog': 9, 'Motorbike': 10, 'People': 11, 'Table': 12}>
	- Labels: [8]
```

and here's the viz:

![image](https://user-images.githubusercontent.com/17111474/116377373-c01cb180-a82e-11eb-9012-95bf92a9d83c.png)

---

### Tracing `draw_sample`

When we start looping through the composites:
https://github.com/airctic/icevision/blob/a95e5e87850fa782b085e5335a7b617bf9b2acaf/icevision/visualize/draw_data.py#L90-L96

we don't check to see if it's a bounding box component or classification component that's being plotted. We simply call `draw_label` inside the loop:
[...]

PR #827, opened 2021-04-28
Draw Sample Improvements

### LINKED ISSUE #785: Dynamic font size for labels
## 🚀 Feature
Dynamically adjust the label font size when displaying records. We can automatically determine the font size based on the width and height of the bbox.

rsomani95:
Addresses #826, #785

* option to explicitly pass x,y coords to `draw_label`
* dynamic font size
* option to include task names
* draw multiple classif tasks neatly

**NOTE:** the dynamic font size logic is quite basic, just does it bases on image height. It still looks ugly when the image width is much wider than the height, thus I don't think we should close #785 with this one.

#### From a custom dataset:

image

#### From `ExDark`:

image

image

### DIFF
--- a/icevision/visualize/draw_data.py
+++ b/icevision/visualize/draw_data.py
@@ -48,6 +48,10 @@ def draw_sample(
     # Args for plotting specific labels
     exclude_labels: List[str] = [],
     include_only: List[str] = None,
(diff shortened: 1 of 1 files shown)


lgvaz (approved):
mergiiing! thanks!

PR #831, opened 2021-05-01
adds iou_thresholds parameter to coco_metric

lgvaz:


icevision/data/convert_records_to_coco_style.py
```diff
@@ -54,7 +58,12 @@ def create_coco_eval(
 
     target_ds = coco_api_from_records(records, show_pbar=show_pbar)
     pred_ds = coco_api_from_preds(preds, show_pbar=show_pbar)
-    return COCOeval(target_ds, pred_ds, metric_type)
+
+    coco_eval = COCOeval(target_ds, pred_ds, metric_type)
+    if iou_thresholds is not None:
+        coco_eval.iouThrs = iou_thresholds
```
kakeros, line 64: Looking at the COCOeval class, I don't think the iouThrs was properly assigned. It should be assigned via `coco_eval.params.iouThrs = iou_thresholds`. I haven't thoroughly tested it, but it should work. However, the logs variable should also be changed accordingly.

PR #833, opened 2021-05-02
adding extra_large yoloV5 backbone

FraPochetti:
I realized YOLOv5 supports a 4th backbone, the `extra_large` one.
Not sure why I didn't notice that in the first place.


lgvaz (approved):
looks gooood, thanks

PR #836, opened 2021-05-04
Updated wandb integration and tutorial.

adamfarquhar:
This works for boundingbox predictions but needs further work to handle masks.


lgvaz:
I change the `__init__` file a bit to be sure the library importing doesn't fail if wandb is not installed, everything else looks good!

PR #848, opened 2021-05-18
Add timm backbones to mmdet models

### LINKED ISSUE #847: Add timm backbones to mmdet models
## 🚀 Feature
timm library has over 250 (pretrained) backbones that could be plugged to mmdet models (over 50 models), and build new object detection / segmentation / keypoints architectures.

ai-fast-track:
I added mmdet + timm support. The API is very similar to the existing one. I separate the namespaces in order to be able to navigate through the different backbones families and have the autocompletion kicking in. 

Both `mobilenet` and `resne(s)t` backbones plugged to a `retinanet` model are provided as an example showing how a user could add support for a specific backbone/model combination. Given we have over 250 timm backbones and over 50 mmet models, it isn’t realistic to imagine a support for all backbone/model combinations. I was thinking about adding a README page that shows how users can add a backbone combination for their use-cases without waiting/adding that to IceVision codebase.

closes #847


icevision/backbones/timm/mobilenet.py
```diff
@@ -0,0 +1,123 @@
...
+#     )
+
+
+# def ice_mobilenetv2_140(pretrained=True, out_indices=(0, 1, 2, 3, 4), **kwargs):
+#     return mobilenetv2_140(
+#         pretrained=pretrained, features_only=True, out_indices=out_indices, **kwargs
+#     )
+
+
+def ice_mobilenetv3_large_075(pretrained=True, out_indices=(0, 1, 2, 3, 4), **kwargs):
```
lgvaz, line 48: Instead of adding `ice` to the model names, can we instead do:

```python
from timm.models import mobilenetv3

def mobilenetv3_large_075(pretrained=True, out_indices=(0, 1, 2, 3, 4), **kwargs):
    return mobilenetv3.mobilenetv3_large_075(...)
```

lgvaz (changes requested):
Seems good on a high level, just a couple of things to address before we merge:

- Carefully handle `timm` imports, right now doing `from icevision.all import *` will try to import `timm` no matter if it's installed or not
- remove commented out code
- rename backbone names that start with `ice`

icevision/backbones/timm/mobilenet.py
ai-fast-track: I removed the timm models from the `icevision/backbones`. It just another indirection and more compelxity

PR #849, opened 2021-05-19
Add PIL Image Compatibility with Record Components

rsomani95:
This is a minimal PR whose aim is just to add the ability to return a `PIL.Image.Image` inside the record components. Backward compatibility is maintained and a `np.ndarray` is returned by default.

This will open the door to `torchvision.transforms` pipelines or `PIL.Image` transforms without additional overhead as we've currently assumed the transforms pipeline to be done using `albumentations`.

### DIFF
--- a/icevision/core/record_components.py
+++ b/icevision/core/record_components.py
@@ -152,16 +152,15 @@ class ImageRecordComponent(RecordComponent):
 
 
 class FilepathRecordComponent(ImageRecordComponent):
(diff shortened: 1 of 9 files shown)


rsomani95:
Side note: Maybe it makes sense to show `Img` instead of `Image` in the repr?
I always look at the record's repr and mistakenly type `record.image` instead of `record.img`

PR #868, opened 2021-06-23
Unet2

lgvaz:
Some introduced "issues":
- because of the new mask workflow, we don't have access to the "number of masks" per annotation anymore (used in autofix)
- `models.get_stats`' `models.get_losses` -> need to find a better namespace

Breaking changes:
- resnet backbone layer cut change

Open issues:
- [ ] test show_results for mask models (torchvision and mmdet); specially for plot_top_losses


FraPochetti:
@lgvaz is this still relevant?

lgvaz:
@FraPochetti Not relevant right? Can be closed I believe

PR #919, opened 2021-09-13
Update installation to torch  1.9 + dependencies

ai-fast-track:
The new installation requirements will avoid reinstalling torch, torchvision and all the other dependencies. Right now, icevision installation lasts more than 5 minutes on Colab because of the re-installation process.


lgvaz (approved):
Looks goood!!

PR #927, opened 2021-09-20
(Update) Updated code to run with albumentations version 1.0.3.

fstroth:
Only a change of the test was required.


setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
     fastcore >=1.3.0,<1.4
     tqdm >=4.49.0,<5
     opencv-python >=4.1.1,<5
-    albumentations >=0.4.6,<0.6
+    albumentations =1.0.3
```
ai-fast-track, line 33: we need to use `==` instead of `=:
```
  albumentations==1.0.3
```

lgvaz (approved):

lgvaz:
Alriight, seems good, that little test failing is not related to this

Thank you!

PR #930, opened 2021-09-22
Fix show pred

potipot:
Since draw_sample_kwargs dict arg passing is missing, we cannot specify font size and other parameters for ground truth plotting. In `show_pred` some of the parameters have defined default values for `draw_pred` but not `draw_sample` which results in different plotting parameters set for records/preds displayed next to each other.
Before:
image

This PR unifies these values adding enforced default font size for both plots.
After:
image


lgvaz (approved):
seems gooood!

lgvaz:
Thank you Paweeeel!

Issue #913, opened 2021-09-01
mask_rcnn.fastai.learner fails with empty annotations


jaeeolma:
## 🐛 Bug
**Describe the bug**

mask_rcnn.fastai.learner fails when it encounters an image with empty mask annotations. I'm not sure whether this is because of icevision, fastai, albumentations or some other thing.

**To Reproduce**
I've encountered two cases where this bug appears: 

1. Have a dataset where some images have no annotations:

```
for r in train_ds.records:
    if len(r.detection.areas) == 0: 
        print(r)

--------------------------------------------------------------------------

BaseRecord

common: 
	- Record ID: 651
	- Filepath: ../data/hiidenportti/processed/7_Hiidenportti_Chunk5_orto/raster_tiles/R0C1.tif
	- Img: None
	- Image size ImgSize(width=500, height=500)
detection: 
	- Areas: []
	- Class Map: <ClassMap: {'background': 0, 'Standing': 1, 'Fallen': 2}>
	- Labels: []
	- BBoxes: []
	- Is Crowds: []
	- Masks: <EncodedRLEs with 0 objects>
...
```

Then depending on where to force the bug to reappear the error message is either

```
---------------------------------------------------------------------------
MemoryError                               Traceback (most recent call last)
<ipython-input-35-69752bdd55ee> in <module>
----> 1 learn.dls.one_batch()

/opt/conda/lib/python3.8/site-packages/fastai/data/load.py in one_batch(self)
    146     def one_batch(self):
    147         if self.n is not None and len(self)==0: raise ValueError(f'This DataLoader does not contain any batches')
--> 148         with self.fake_l.no_multiproc(): res = first(self)
    149         if hasattr(self, 'it'): delattr(self, 'it')
    150         return res

/opt/conda/lib/python3.8/site-packages/fastcore/basics.py in first(x, f, negate, **kwargs)
    545     x = iter(x)
    546     if f: x = filter_ex(x, f=f, negate=negate, gen=True, **kwargs)
--> 547     return next(x, None)
    548 
    549 # Cell

/opt/conda/lib/python3.8/site-packages/fastai/data/load.py in __iter__(self)
    107         self.before_iter()
    108         self.__idxs=self.get_idxs() # called in context of main process (not workers/subprocesses)
--> 109         for b in _loaders[self.fake_l.num_workers==0](self.fake_l):
    110             if self.device is not None: b = to_device(b, self.device)
    111             yield self.after_batch(b)

[...]

PR #933, opened 2021-09-24
Empty mask fix

jaeeolma:
Related to #913, fixes to the following:

* Force empty mask annotations to have valid shape with quick hack.
  * This was breaking `COCOMetric(metric_type=COCOMetricType.mask)`
* Avoid trying to concatenate empty masks
  * This was breaking a lot of things
* Just to be sure add check for empty `self._keep_mask` list in `albumentations_adapter/_filter_attribute`
  * This might not be needed anymore. Seems that original problem was related to faulty masks but at least I found some other bugs to squish

### DIFF
--- a/icevision/core/mask.py
+++ b/icevision/core/mask.py
@@ -82,9 +82,17 @@ class MaskArray(Mask):
         return self
 
     def to_erles(self, h, w) -> EncodedRLEs:
-        return EncodedRLEs(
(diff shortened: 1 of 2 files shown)


ai-fast-track:
Thanks a lot for your contribution @jaeeolma!

Sorry for the long delay.

PR #934, opened 2021-09-25
Initial mmsegmentation support

nicjac:
Initial `mmsegmentation` support. This is a work-in-progress.

Will populate this PR as we go!

- [x] Integration can be used end-to-end with **DeepLabV3** and **SegFormer**
- [x] W&B integration for semantic segmentation
- [x] New checkpoint export/import support (this one might have to wait as feature was added after the fork)
- [x] Custom CFG like mmdet implementation (might already be done, but should be tested)
- [x] Properly implement "hack" to avoid Pytorch doing parallel training by default in utils.py
- [x] Interpretation loop (currently using `mmdet`'s)
- [x] Updated semantic segmentation notebook to showcase `mmsegmentation` models

To figure out:
- [x] Splitter / Trainable Params / param_groups
- [x] ~~Could we use fastai's Dice and other metrics (if fastai is a hard dependency)?~~ Will keep current implementation

To add in an upcoming PR:
- [ ] Lightning integration
- [ ] CPU support
- [ ] Tests for `mmsegmentation` models


lgvaz:
Heeey nicjac! This is a big and a very very helpful PR, we'll try to assist you as much as we can, feel free to ask for help at any point =)

FraPochetti:
Hi @nicjac any update here?
cc @ai-fast-track @fstroth @potipot

nicjac:
> Hi @nicjac any update here? cc @ai-fast-track @fstroth @potipot

Yep! We've been using this implementation internally for a while, but got sidetracked by other priorities.

Work has resumed, will update the PR and make sure it can be merged.

nicjac:
This looks good to go to me!

icevision/models/mmseg/lightning/model_adapter.py
```diff
@@ -0,0 +1,67 @@
...
+        preds = self.convert_raw_predictions(
+            batch=data, raw_preds=raw_preds, records=records
+        )
+        self.accumulate_metrics(preds)
+
+        for k, v in outputs["log_vars"].items():
+            self.log(f"valid/{k}", v)
+
+        # TODO: is train and eval model automatically set by lighnting?
+        self.model.train()
```
FraPochetti, line 64: @potipot this is not needed right?
As @nicjac says PL should handle the `.train()`/`.eval()` switch

icevision/models/mmseg/lightning/model_adapter.py
```diff
@@ -0,0 +1,67 @@
...
+    # Returns
+        A `LightningModule`.
+    """
+
+    def __init__(self, model: nn.Module, metrics: List[Metric] = None):
+        super().__init__(metrics=metrics)
+        self.model = model
+
+    @abstractmethod
+    def convert_raw_predictions(self, batch, raw_preds, records):
```
FraPochetti, line 30: pure python question. How can this work without a `return` or `pass` statement inside the function?

FraPochetti: What function gets invoked [here](https://github.com/airctic/icevision/pull/934/files#diff-b52b136bda8d73b70540f7af32ce576d54b6644eb6c25ba4f08ee46df367a760R55) exactly?

FraPochetti:
I cannot preview the `notebooks/getting_started_semantic_segmentation.ipynb` notebook.
Is [this](https://colab.research.google.com/github/Orbis-International/icevision/blob/mmseg-initial-support/notebooks/getting_started_semantic_segmentation.ipynb) how it looks like?

icevision/models/mmseg/models/deeplabv3plus/backbones/resnet.py
```diff
@@ -0,0 +1,283 @@
...
+base_config_path = mmseg_configs_path / f"{model_name}"
+base_weights_url = "https://download.openmmlab.com/mmsegmentation/v0.5/deeplabv3plus"
+
+resnet101_d16_mg124 = MMSegDeeplabPlusBackboneConfig(
+    backbone_type="R101-D16-MG124",
+    pre_trained_variants=[
+        {
+            "config_path": "deeplabv3plus/deeplabv3plus_r101-d16-mg124_512x1024_40k_cityscapes.py",
+            "crop_size": (512, 1024),
+            "lr_schd": "40k",
```
FraPochetti, line 31: what is `lr_schd`?

icevision/metrics/segmentation_accuracy/segmentation_accuracy.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Arguments
+        ignore_class: class to ignore during metric computation (e.g. void class).
+
+    """
+
+    _seg_masks_gt = None
+    _seg_masks_pred = None
+
+    def __init__(self, ignore_class=None):
```
FraPochetti, line 22: Why don't we set this to `Void` as default?

FraPochetti:
Out of curiosity, how do you handle the end2end inference flow with masks (e.g. return the mask on the original image, not the transformed one)?
The current implementation of `end2end_detect` ([here](https://github.com/airctic/icevision/blob/4f68a87baa6651766f406b7cc9bf8d377dc98d20/icevision/models/inference.py#L18)) supports only `bboxes`.

icevision/models/mmseg/README.md
```diff
@@ -0,0 +1,29 @@
...
+# How the integration works
+
+The integration is very similar to that of `mmdet` (as both libraries share much of their structure).
+
+There are multiple layers of integrations, which are described here to facilitate future developments. 
+
+## Callbacks
+
+The [`MMSegmentationCallback`](icevision/models/mmseg/fastai/callbacks.py) is responsible for making the `mmsegmentation` models compatible with the fastai training loop. It ensures that:
+- The `after_create` hook is ued to wrap the model in a way that `mmsegmentation`'s conventions are translated to fit with fastai's model operations
```
FraPochetti, line 14: Small typo.
`ued` instead of `used`

icevision/core/mask.py
```diff
@@ -64,10 +64,11 @@ class MaskArray(Mask):
 
     # Arguments
         data: Mask array, with the dimensions: (num_instances, height, width)
+        pad_dim: bool
     """
 
-    def __init__(self, data: np.uint8):
-        if len(data.shape) == 2:
+    def __init__(self, data: np.uint8, pad_dim: bool = True):
+        if pad_dim and (len(data.shape) == 2):
```
FraPochetti, line 71: What is `pad_dim` for?

icevision/models/fastai/unet/fastai/callbacks.py
```diff
@@ -10,3 +11,15 @@ def before_batch(self):
         self.learn.records = self.yb[0]
         self.learn.yb = (self.xb[0][1],)
         self.learn.xb = (self.xb[0][0],)
+
+    def after_pred(self):
```
FraPochetti, line 15: Why was this one added?
Was there something broken with the UNet implementation?

icevision/models/fastai/unet/prediction.py
```diff
@@ -39,10 +39,21 @@ def convert_raw_predictions(
     keep_images: bool = False,
 ) -> List[Prediction]:
     mask_preds = raw_preds.argmax(dim=1)
-    tensor_images, *_ = batch
```
FraPochetti, line 42: Same comment as the previous file.
UNet is outside of `mmseg`.
Did anything break because of the changes to `MaskArray`, hence you were obliged to edit those files?

icevision/models/fastai/unet/prediction.py
```diff
@@ -39,10 +39,21 @@ def convert_raw_predictions(
...
+
+    if len(batch) > 1:
+        tensor_images, tensor_gts = batch
+    else:
```
FraPochetti, line 45: Does this happen at inference?
E.g. when there are no masks?

icevision/models/interpretation.py
```diff
@@ -223,7 +224,7 @@ def plot_top_losses(
         samples = add_annotations(samples)
 
         dl = self.infer_dl(dataset, batch_size=batch_size)
-        preds = self.predict_from_dl(model=model, infer_dl=dl)
+        preds = self.predict_from_dl(model=model, infer_dl=dl, keep_images=True)
```
FraPochetti, line 227: Why do we need `keep_images` here?
We rebuild `Prediction` objects from `sorted_samples` [here](https://github.com/airctic/icevision/pull/934/files#diff-f3f81ce939d27f5cf5ed326cf50fc5cf26d47b9d1cf92ac04a22d45c802b4807R256) which should have `img`s set properly already.

tests/models/mmseg/test_show_results.py
```diff
@@ -0,0 +1,63 @@
+import pytest
+from icevision.all import *
+from icevision.models.mmseg.models.deeplabv3.backbones import *
+from icevision.models.mmseg.models import *
+from icevision.models.interpretation import get_samples_losses
+
```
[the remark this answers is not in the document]
nicjac: I agree! I kind of ran out of time to implement more tests. Happy to do some (either for this PR, or a future refinement pass).

icevision_install.sh
```diff
@@ -58,6 +61,14 @@ case ${icevision_version} in
       pip install git+https://github.com/airctic/icedata.git --upgrade -q
       ;;
 
+   mmseg-integration)
+      echo "- Installing icevision from master (icevision integration)"
+      pip install git+https://github.com/Orbis-International/icevision.git@05ad41f183132275c7f31f25e484fcbff0a7cfd8#egg=icevision[all] --upgrade -q
```
FraPochetti, line 66: This line needs to be edited before merging

icevision/models/mmseg/models/deeplabv3plus/backbones/resnet.py
nicjac: It corresponds to different learning schedules used during the training of pretrained models (see https://github.com/open-mmlab/mmsegmentation/tree/master/configs/_base_/schedules).

This is one of the many complexities of `mmsegmentation` - there are many variations and permutations when it comes to pretrained models: different datasets, different learning schedules, different input size, etc...

icevision/metrics/segmentation_accuracy/segmentation_accuracy.py
nicjac: This is an arbitrary choice on my part - I thought the expected default behaviour would be not to ignore any class, though I can see why `Void` would make sense here. Happy either way.

icevision/models/fastai/unet/fastai/callbacks.py
nicjac: Indeed - the original `unet` implementation did not convert predictions to the "standard" IceVision format and was carrying out operations using the raw output format. I think this was fine before as it was kind of a stand-out model being the only way to do semantic segmentation. However, it meant that it was not compatible with everything else being built for semantic seg (in this PR and elsewhere).

For example, it was not possible to have both `unet` and `mmsegmentation` models in the same tutorial notebook as `unet` would fail with the newly introduced semantic segmentation metrics.

icevision/core/mask.py
nicjac: This change introduced the option **not to** pad the mask dimensions. Before this change, the mask array always ended up being padded.

icevision/models/mmseg/README.md
nicjac: Thanks 👍

icevision/metrics/segmentation_accuracy/segmentation_accuracy.py
FraPochetti: @potipot @fstroth @ai-fast-track wdyt?

icevision/models/interpretation.py
nicjac: I can't double check right now but I _think_ this is related to the fact that without `keep_images` set to `True`, GT masks would not be available for subsequent operations. Happy to revisit this one if you think it is needed.

tests/models/mmseg/test_show_results.py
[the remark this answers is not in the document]
FraPochetti: I'd be more comfortable with a couple more added to this PR.
Doesn't need to be anything great.
You can totally copy the ones we already have for `mmdet`.

icevision/core/mask.py
FraPochetti: During the transforms?

tests/models/mmseg/test_show_results.py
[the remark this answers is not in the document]
FraPochetti: @ai-fast-track ?

icevision/models/fastai/unet/prediction.py
nicjac: At inference time without GT masks being available, the second condition is triggered (and `tensor_gts` is set as an array of `None`).

This is admittedly a bit hacky, but was in still in an effort to get `unet` working in the more general semantic segmentation framework.

nicjac:
> Out of curiosity, how do you handle the end2end inference flow with masks (e.g. return the mask on the original image, not the transformed one)? The current implementation of `end2end_detect` ([here](https://github.com/airctic/icevision/blob/4f68a87baa6651766f406b7cc9bf8d377dc98d20/icevision/models/inference.py#L18)) supports only `bboxes`.

Unfortunately, it does not just yet. I have a _very_ hacky implementation for prototyping purpose but a proper implementation will be part of a refinement PR. It shouldn't take too long as we do need this for our work as well!

icevision/models/interpretation.py
FraPochetti: You are right that the masks would not be available in GT, given the records get unloaded.
I think this is not a problem specific to the `plot_top_losses` functionality as we get GT records from [here](https://github.com/airctic/icevision/pull/934/files/6de58ba0fcd3f1af7d9c6981522b68615fe92d8c#diff-f3f81ce939d27f5cf5ed326cf50fc5cf26d47b9d1cf92ac04a22d45c802b4807R223). We don't take them from the `preds` object. As you can see [here](https://github.com/airctic/icevision/pull/934/files/6de58ba0fcd3f1af7d9c6981522b68615fe92d8c#diff-f3f81ce939d27f5cf5ed326cf50fc5cf26d47b9d1cf92ac04a22d45c802b4807R228) from `preds` we get only the predictions. Long story short the GT records are the same we had ad inputs in the dataset, therefore they are not unloaded. They have everything set properly.

nicjac: I remember adding this due to tests related to `plot_top_losses` failing - maybe it isn't actually needed?

FraPochetti: I'd be surprised if it was needed. Can you check removing it and see what happens?

FraPochetti:
> > Out of curiosity, how do you handle the end2end inference flow with masks (e.g. return the mask on the original image, not the transformed one)? The current implementation of `end2end_detect` ([here](https://github.com/airctic/icevision/blob/4f68a87baa6651766f406b7cc9bf8d377dc98d20/icevision/models/inference.py#L18)) supports only `bboxes`.
> 
> Unfortunately, it does not just yet. I have a _very_ hacky implementation for prototyping purpose but a proper implementation will be part of a refinement PR. It shouldn't take too long as we do need this for our work as well!

I am very interested in this one, so I can definitely take care of it too.

tests/models/mmseg/test_show_results.py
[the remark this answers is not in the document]
ai-fast-track: @FraPochetti  @nicjac I agree we need to have at least some basic tests to make sure nothing breaks. We can add more tests in subsequent PRs.

icevision/models/mmseg/lightning/model_adapter.py
potipot: yes, every validation step is vrapped with `.train()` call by pytorch-lightning https://pytorch-lightning.readthedocs.io/en/latest/common/lightning_module.html#validation

FraPochetti: @nicjac I guess we can remove the line, then?

icevision/models/mmseg/README.md
FraPochetti: This is not addressed yet, right?

nicjac: It is now :)

icevision_install.sh
nicjac: Forgot this was left in, now removed

icevision/models/interpretation.py
nicjac: I removed it, and tests still work. Doesn't seem to be break anything, so will keep it like that for now!

FraPochetti: Thanks for checking man!

icevision/metrics/segmentation_accuracy/segmentation_accuracy.py
potipot: I'm not familiar with and never used `Void`. In context like this I also usually set var to None

FraPochetti: @lgvaz what are your 2 cents here?

potipot: On another note though, why is this set as class attribute rather than instance, set on init?

FraPochetti: Does it matter?

potipot: My intuition was right, take a look at this example:
image

FraPochetti: Wait, Ia m referring to `ignore_class` which is set as an instance attribute, not a class one.

potipot: Oh, sorry I totally misunderstood the issue then.

FraPochetti: Np man.
When I asked
> Why don't we set this to Void as default?

I meant 
"Why don't we set this ignore_class to Void as default?"

FraPochetti: Thanks for the discussion guys.
Thinking about it more thoroughly, I am starting to convince myself that it could be a good idea to set `ignore_class=None` as it's done now. I am not sure but I think that `Void` is CamVid specific.

PR #960, opened 2021-10-30
(Feature) Added way to output metrics for pytorchlightning during training

### LINKED ISSUE #600: Add more logging to the pytorch lighning models.
The feature consists of two parts:
  1. Add the validation loss to the progress bar by default
  2. Create boolean parameter for extended progress bar logging (showing the different components of the loss)

fstroth:
With this PR the `LightningModelAdapter` will take another `__init__` argument named `metrics_keys_to_log_to_prog_bar` which has to be a list of string tuples with two elements, were the first entry is the name of the metric to log and the second one is the display name.
By default the value will be set to `[("AP (IoU=0.50:0.95) area=all", "mAP")]` which will log the mAP. This will only log the value is the COCOMetric is supplied as a metric! If so it will look like this:

image

Closes #600

### DIFF
--- a/icevision/engines/lightning/lightning_model_adapter.py
+++ b/icevision/engines/lightning/lightning_model_adapter.py
@@ -19,7 +19,7 @@ class LightningModelAdapter(pl.LightningModule, ABC):
         super().__init__()
         self.metrics = metrics or []
         self.metrics_keys_to_log_to_prog_bar = metrics_keys_to_log_to_prog_bar or [
-            ("AP (IoU=0.50:0.95) area=all", "mAP")
+            ("AP (IoU=0.50:0.95) area=all", "COCOMetric")
         ]
 
     def accumulate_metrics(self, preds):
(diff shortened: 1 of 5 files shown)


fstroth:
The test seem to fail because `pytest-mock` is not installed for the test environment. Where should I add the dependency?

ai-fast-track:
That's great @fstroth!

I think it would be good to use `COCOMetric` as the default title to be consistent with our current Fastai logging:
instead of: 
`[("AP (IoU=0.50:0.95) area=all", "mAP")]`
We would use this:
`[("AP (IoU=0.50:0.95) area=all", "COCOMetric")]

image

ai-fast-track:
> The test seem to fail because `pytest-mock` is not installed for the test environment. Where should I add the dependency?

In the `setup.cfg`  `[dev]` section?
https://github.com/airctic/icevision/blob/master/setup.cfg#L56

fstroth:
@ai-fast-track Yes the test is passing locally, all other tests also pass except the yolo ones because I didn't install the yolo part. Training a model with the code also worked (that's where the screenshot above is from).

PR #963, opened 2021-10-31
(Feature) init coco and via parser with a dict instead of the filepath

fstroth:
With this PR the filepath argument for the coco and via parser can be a dict. This is usefull if the annotations are not loaded from disk but from an api.
The name of the init parametr filepath was not changed to not break the api.

### DIFF
--- a/icevision/parsers/coco_parser.py
+++ b/icevision/parsers/coco_parser.py
@@ -31,12 +31,15 @@ def coco(
 class COCOBaseParser(Parser):
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


ai-fast-track:
> With this PR the filepath argument for the coco and via parser can be a dict. This is usefull if the annotations are not loaded from disk but from an api. The name of the init parametr filepath was not changed to not break the api.

Thanks for the PR @fstroth. Is this use-case found in the IceVision Dashboard?

PR #965, opened 2021-11-02
fix image IO

Borda:
Fix for using old PIL library which is failing for:
```
AttributeError: module 'PIL.TiffTags' has no attribute 'IFD'
```

### DIFF
--- a/icevision/utils/imageio.py
+++ b/icevision/utils/imageio.py
@@ -59,8 +59,8 @@ def get_img_size(filepath: Union[str, Path]) -> ImgSize:
         with PIL.Image.open(filepath) as image:
             image_size = image.size
     except AttributeError as err:
-        im = torchvision.io.read_image(filepath)
-        image_size = tuple(im.shape[-2:][::-1])
+        im = cv2.imread(filepath)
+        image_size = tuple(im.shape[:2][::-1])
 
     try:
         exif = image._getexif()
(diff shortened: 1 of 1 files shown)


Borda:
@lgvaz @ai-fast-track could you please approve CI run? :rabbit:

Borda:
note that it keeps using PIL as a primary tool, and ties to use OpenCV only if the PIL tries fails...

Borda:
for the record, there are not so many loading alternatives...
https://towardsdatascience.com/what-library-can-load-image-in-python-and-what-are-their-difference-d1628c6623ad

Borda:
yeah, the codecov is expected to fail as it would be required to create such a corrupted image file and upload it to git for testing...

Borda:
@lgvaz any update here? :rabbit:

Borda:
any update here? @FraPochetti

FraPochetti:
Hi @Borda , 
first of all please accept our apologies for responding so late. 
This issue had fallen through the cracks and it got just buried among notifications.

We have discussed your PR and I am afraid we cannot accept it.
1. Nobody else has raised this specific issue except you, and we are not able to reproduce it.
2. We have worked hard to move out of `opencv` to handle images, and transitioned entirely to PIL a while back. Even if you are proposing just a fallback option to catch the error, we prefer to stick to our strategy.

Thanks for your understanding.

Issue #987, opened 2021-12-08
Add detection_threshold arg for all Lightning adapters


Anjum48:
## 🚀 Feature
**Is your feature request related to a problem? Please describe.**
At the moment, the detection threshold for inside `validation_step` is at or near zero for Lightning adapters. This results in pessimistic validation performance due to false positives.
https://github.com/airctic/icevision/blob/9abbb09f47e6fc18867a5aa662885b558af2975c/icevision/models/ross/efficientdet/lightning/model_adapter.py#L50
https://github.com/airctic/icevision/blob/9abbb09f47e6fc18867a5aa662885b558af2975c/icevision/models/ultralytics/yolov5/lightning/model_adapter.py#L51

**Describe the solution you'd like**
Add a kwarg in the init of the Lightning adapter

**Describe alternatives you've considered**

**Additional context**
Happy to submit a PR if you guys agree

FraPochetti:
@potipot @fstroth @ai-fast-track wdyt?

fstroth:
Sounds good to me, we should go forward with this.

FraPochetti:
@Anjum48 would you like to submit a PR?
We'd love that!

potipot:
Ah, this is actually the way it should be. When calculating COCOMetric (or mAP), we don't put the confidence threshold on detections but keep the best 100 of them for each image.

FraPochetti:
Ah wait, so are you saying that we should keep the threshold to 0?
I think this doesn't change in the [PR](https://github.com/airctic/icevision/pull/1004), but can you take a look too?

Anjum48:
Thanks for the clarification @potipot! I guess I was coming from the angle that a custom metric may need a confidence threshold, hence the need for this (in my case I implemented the F2 score from [this](https://www.kaggle.com/c/tensorflow-great-barrier-reef/overview/evaluation) ongoing competition.

I guess there might be some complexity in the future if you are using something like:
```
metrics = [COCOMetric(), F2Score()]
```
where `COCOMetric` does not require a threshold but `F2Score` does. In this particular case, would it be more sensible to put the threshold within the custom metric also?

potipot:
yes exactly, I think it should be up to the metric to handle the predictions. 

[...]

PR #1004, opened 2021-12-19
Issue 987 - Add detection_threshold arg for all Lightning adapters

Anjum48:
Fix for issue #987. Any feedback is appreciated! 😀️

The only bit I wasn't sure about was the `convert_raw_predictions` in the `mmdet` adapter (the `import *` style makes tracing dependence tricky). I had to remove `self.` to make it work but I'm not sure if that would break other things.


FraPochetti:
Hi @Anjum48, thanks for your contribution!
I see some tests are failing.
Is everything green on your local machine?

Anjum48:
Hi @FraPochetti, Yes I'm seeing failed tests locally too. I'll try and track them down and get back to you if I need some help

Anjum48:
All tests passing on my side so 🤞️

FraPochetti:
Nice! All tests are passing. We will review ASAP.

icevision/models/mmdet/lightning/model_adapter.py
```diff
@@ -59,16 +66,16 @@ def validation_step(self, batch, batch_idx):
...
+            records=records,
+            detection_threshold=self.detection_threshold,
         )
         self.accumulate_metrics(preds)
 
         for k, v in outputs["log_vars"].items():
             self.log(f"valid/{k}", v)
 
-        # TODO: is train and eval model automatically set by lighnting?
-        self.model.train()
```
FraPochetti, line 71: Why did you remove this line?

Anjum48: PyTorch Lightning automatically takes care of `model.train()` and `model.eval()`. I removed it to make it consistent with the other lightning adapters

fstroth:
Looks all good to me! The only suggestion I would have would be to change the parameter name to `validation_detection_threshold` to indicate the intention/usage of the threshold. Should make it easier to understand the code when reading it for the first time or when people check the init parameters.

icevision/models/mmdet/lightning/model_adapter.py
FraPochetti: @potipot do you recall why this was there in the first place?
I think @Anjum48 point is valid.

FraPochetti:
> Looks all good to me! The only suggestion I would have would be to change the parameter name to `validation_detection_threshold` to indicate the intention/usage of the threshold. Should make it easier to understand the code when reading it for the first time or when people check the init parameters.

@Anjum48 would you mind implementing @fstroth suggestion, please?

potipot:
I think `validation_confidence_threshold` might be even more accuratem WDYT?

FraPochetti:
> I think `validation_confidence_threshold` might be even more accuratem WDYT?

+ 1

PR #1005, opened 2021-12-20
Fix checkpoint loading when model_name contains models or additional components

gablanouette:
Updates to the model_from_checkpoint function to ensure handling of efficientdet based on various ways to name models

### DIFF
--- a/icevision/models/checkpoint.py
+++ b/icevision/models/checkpoint.py
@@ -206,8 +206,10 @@ def model_from_checkpoint(
     model_type = None
     if model_name:
         model = model_name.split(".")
+        # If model_name contains three or more components, the library and model are the second to last and last components, respectively (e.g. models.mmdet.retinanet)
         if len(model) >= 3:
             model_type = getattr(getattr(models, model[-2]), model[-1])
+        # If model_name follows the default convention, the library and model are the first and second components, respectively (e.g. mmdet.retinanet)
         else:
             model_type = getattr(getattr(models, model[0]), model[1])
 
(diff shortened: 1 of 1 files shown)


gablanouette:
updates to model_from_checkpoint function to be able to handle efficientdet models

FraPochetti:
Hi @gablanouette I already see that the formatting check is failing.
You need to run `black`.
You can do that automatically using `pre-commit hooks`.
Check them [here](https://airctic.com/0.11.0/contributing/).

icevision/models/checkpoint.py
```diff
@@ -208,8 +205,11 @@ def model_from_checkpoint(
 
     model_type = None
     if model_name:
-        lib, mod = model_name.split(".")
```
FraPochetti, line 211: I see `lib` was instantiated before but it's not instantiated anymore in your PR.
Was it used somewhere else?
Any reason behind its removal?

gablanouette: It's no longer a binary splits so i do not believe they are needed any longer

FraPochetti:
Pending https://discord.com/channels/735877944085446747/735877944517591151/922895633982062682

gablanouette:
The code has been updated to add comments explaining why this is necessary to the users. For each case, if the length of the components in model_name is greater than 3, (e.g. models.mmdet.retinanet), we only want to use the library and model which are the second to last and last components. Otherwise, if model_name follows the default convention, the library and model are the first and second components, respectively (e.g. mmdet.retinanet)

ai-fast-track:
Thanks a lot for your first contribution, @gablanouette! 🎉🥳

PR #1008, opened 2021-12-22
fixing the OCHuman notebook

FraPochetti:


potipot:
Going about removing extra notebooks, what about `ochuman_keypoint_detection.ipynb` vs `getting_started_keypoint_detection.ipynb` . What's the motivation for keeping two of them?

FraPochetti:
Good question.
BIWI is a super simple dataset and it doesn't really show the flexibility and power of Keypoints detection.
But we need some easy to access dataset to show the functionality. So we kept it.
OCHuman is a lot more complicated and impressive. It gets the message across better.
So BIWI answers the question "how does  this work?" whereas OCHuman answers "What can I actually do with it?"

PR #1013, opened 2021-12-31
fixing the negative_samples notebook

potipot:
- [x] added recordcollection add operator (+)
- [x] add slicing (slice indexing) for RecordCollection to return a new object of the same class
- [x] fix the negative samples notebook
- [ ] add tests for the new operator
- [x] fix failing tests


FraPochetti:
A couple of tests are not passing apparently.

FraPochetti:
why are the changes to `RecordCollection` needed?

potipot:
These changes allow more intuitive use of the old records list, a bit like regular lists. This slicing and adding is used in the negative samples notebook, but I also came across this use case and already had these commits in my own branch done some time before.  

The idea is to be able to 
- do RecordCollection + RecordCollection 
- and make sure that RecordCollection[:n] returns a new RecordCollection instance (and not List[BaseRecord]), cause then we cannot do RecordCollection + RecordCollection[:10] (Operator + not defined for RC and List)

## TODO;
You;re right I hurried too much with the PR. I will add tests to the new operator and fix the ones failing.

potipot:
the failing tests are the known efficientdet metric values

fstroth:
Looks good to me. @FraPochetti  I think we can merge.

FraPochetti:
> Looks good to me. @FraPochetti I think we can merge.

Yes! Let's merge!

FraPochetti (approved):

PR #1022, opened 2022-01-10
fix efficientdet metrics failing test

potipot:
- [x] update torch and mmcv versions in github python environment
- [x] recreate github env on local machine
- [x] identify the source of the problem - different Pillow version
- [x] fix Pillow version requirements
- [x] github action testing environment setup from `icevision_install.sh` instead of explicit `pip commands`
- [x] add yolov5-icevision and sahi to [all] dependencies 
- [x] add tests for RecordCollection to satisfy codecov

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,7 +59,7 @@ dev =
   keras-autodoc ==0.6.0
   mkdocs >=1.1.2,<2
   mkdocs-material >=7.0.6,<8
-  mike >=0.5.5,<1
+  mike >=1.0
   jupyter >=1.0.0,<2
   pymdown-extensions >=8.0,<9
   Sphinx >=3.1.0,<4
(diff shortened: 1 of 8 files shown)


setup.cfg
```diff
@@ -59,7 +61,7 @@ dev =
   keras-autodoc ==0.6.0
   mkdocs >=1.1.2,<2
   mkdocs-material >=7.0.6,<8
-  mike >=0.5.5,<1
+  mike >=1.0
```
potipot, line 64: this is updated according to a comment left:
```
 # Below version is needed to work around a clash with ruamel-yaml. When v1.0 is released this can be removed.
pip install git+git://github.com/jimporter/mike.git@2961733cbfef132e9d63752307443981ad02d1fc --upgrade
```
the issue was that `mkdocs.yml` couldnt be declaring the same keyword multiple times. From 1.0 it seems to be working.

FraPochetti (approved):

PR #1029, opened 2022-01-18
refactor: Fixed the bug to add neck modules properly

hectorLop:
Fix a bug introduced by the DETR implementation. The problem was that when iterating the `named_modules()`, it return both the name and the module, but it was iterated as it was a list so the statement  `if name == 'neck'` was never going to be True.

### DIFF
--- a/icevision/models/mmdet/utils.py
+++ b/icevision/models/mmdet/utils.py
@@ -47,7 +47,10 @@ def param_groups(model):
         layers += [getattr(body, l) for l in body.res_layers]
 
     # add the neck module if it exists (DETR doesn't have a neck module)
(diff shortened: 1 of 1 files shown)


hectorLop:
Here is a [colab](https://colab.research.google.com/drive/147-5CGrIPFJKHBaO5dk9gZ6OLiNz6I6D?usp=sharing) notebook, similar to the getting started, but I added a section where I create the YOLOx, Retinanet and DETR models and use fastai to find their optimar learning rates. Thus, it shows that now they can be created succesfully.

FraPochetti:
Thanks a ton again!

PR #1031, opened 2022-01-18
(feat) Added utils to convert Records and Predictions to fiftyone

2649:
The first draft for an FO integration. I mimicked the format of `convert_records_to_coco_style.py` and added it under the `icevision.data ` namespace. 

Currently, it supports only bboxes, but we can extend it easily to other formats. 

It is designed to use the icevison types Records and Predictions. 

@ai-fast-track Maybe you could help me with adding fiftyone as new dependency. I'm not quite sure, what is the recommended way. Maybe, we do some kind of lazy import like with mmdet, because the fiftyone is quite heavy, it uses a mongDB under the hood and is for linux users only.


FraPochetti:
Awesome work man!
Thanks a lot for contributing!

ai-fast-track:
Thanks a lot, Laurenz!

ai-fast-track:
Can you add a `pip install fiftyone` 

- in the `build-pkg.yml` installation section here below ():
https://github.com/airctic/icevision/blob/4f626f4aa723be13a1b8665dc56dd3b1aeb0f439/.github/workflows/build-pkg.yml#L26

- and in the `mk-docs-build.yml` docs installation section here :
https://github.com/airctic/icevision/blob/4f626f4aa723be13a1b8665dc56dd3b1aeb0f439/.github/workflows/mk-docs-build.yml#L32

ai-fast-track:
Can you please add `pip install fiftyone`

- in `build-pkg.yml`
https://github.com/airctic/icevision/blob/4f626f4aa723be13a1b8665dc56dd3b1aeb0f439/.github/workflows/build-pkg.yml#L26

- and in `mk-docs-build.yml`
https://github.com/airctic/icevision/blob/4f626f4aa723be13a1b8665dc56dd3b1aeb0f439/.github/workflows/mk-docs-build.yml#L30

FraPochetti:
I think it would be great to also have notebook showcasing the integration.
Ideally as part of this PR.
Something we will add to the `notebooks` folder and to the docs

ai-fast-track:
Thanks a lot @2649 for adding such a detailed walk-through notebook. Amazing Job!

PR #1033, opened 2022-01-19
Add Swin Transformer backbone for Mask RCNN

dnth:
This PR adds the Swin Transformer backbone for Mask RCNN.


dnth:
Is there a way to use this backbone for object detection?

FraPochetti:
Merging! Thanks a lot man!
Keep rocking!

FraPochetti:
>  Is there a way to use this backbone for object detection?

Not sure actually

PR #1050, opened 2022-01-27
CentripetalNet Support

dnth:
This PR adds CentripetalNet with HourglassNet backbone to IceVision. Model only works for object detection task. No mask related config found on the mmdet [repo](https://github.com/open-mmlab/mmdetection/tree/master/configs/centripetalnet) for instance segmentation task.
The following screenshot shows COCOMetric during training using the getting started notebook on my local machine.

image

Also sharing the training on Colab
https://colab.research.google.com/drive/1ADQ8mts1GRc-_y1CqGMvWfZ7ipK4LfWV?usp=sharing


icevision/models/mmdet/utils.py
```diff
@@ -44,6 +45,17 @@ def param_groups(model):
...
+        layers += [
+            body.stem,
+            body.hourglass_modules,
+            body.inters,
+            body.conv1x1s,
+            body.out_convs,
+            body.remap_convs,
+            body.relu,
+        ]
+
```
ai-fast-track, line 58: This is unrelated to the CentripetalNet model but I wanted to clarify the `param_groups` for the YOLACT model section here below:

Should it be `True` instead of `False` in the `if` statements?
``` 
        # YOLACT has mask_head and segm_head
        if getattr(model, "mask_head", False):
            layers += [model.mask_head]
        if getattr(model, "segm_head", False):
            layers += [model.segm_head]
```

dnth: Thank you for the query. From what I understood, the last argument in the `getattr` function is the default value to return if the attribute does not exist. I referred to some examples [here](https://www.w3schools.com/python/ref_func_getattr.asp).

So, in `getattr(model, "mask_head", False)` it means that the function will return `False` if the `mask_head` attribute does not exist.

I'm not sure if there is a better way to check this attribute so its less confusing?

dnth: I notice that these works too

```
        # YOLACT has mask_head and segm_head
        if getattr(model, "mask_head"):
            layers += [model.mask_head]
        if getattr(model, "segm_head"):
            layers += [model.segm_head]
```

ai-fast-track: I see! Thanks!

dnth: I'm sorry, I wasnt very thorough in testing. We will need `getattr(model, "mask_head", False)` after all. Otherwise the getting started notebook fails https://github.com/airctic/icevision/issues/1057

PR #1055, opened 2022-02-01
removing fiftyone as a hard dependency

FraPochetti:
* Removed `fiftyone` as a hard dependency. Needs to be installed separately now.
* Added disclaimer to `fiftyone` notebook about Colab incompatibility.
* Made sure every notebook installs from pypi and not from master.


ai-fast-track:
Thanks a lot, @FraPochetti!

Zubair2019:
> * Removed `fiftyone` as a hard dependency. Needs to be installed separately now.
> * Added disclaimer to `fiftyone` notebook about Colab incompatibility.
> * Made sure every notebook installs from pypi and not from master.

Have you checked which model are added when installing via PyPi. Last time I checked their just a few models in the mmdet directory. Please check.

PR #1056, opened 2022-02-03
Adding Coco Metric for inference

ai-fast-track:
I added some convenient methods to easily calculate the COCO metric at inference when ground truth records are available

```
infer_dl = model_type.infer_dl(valid_ds, batch_size=4, shuffle=False)
preds = model_type.predict_from_dl(model, infer_dl, keep_images=True)
```
```
inference_metric = COCOMetric(metric_type=COCOMetricType.bbox)
```
Get the COCO metric from the Predictions object:
```
logs = inference_metric.metric_from_preds(preds, print_summary=True)
```

You call also separately pass the ground_truth and the predictions:
```
gt_list = [pred.ground_truth for pred in preds]
preds_list = [pred.pred for pred in preds]

logs = inference_metric.metric(gt_list, preds_list, print_summary=True)
```

### DIFF
--- a/icevision/models/mmdet/utils.py
+++ b/icevision/models/mmdet/utils.py
@@ -81,9 +81,9 @@ def param_groups(model):
         layers += [model.bbox_head]
 
         # YOLACT has mask_head and segm_head
-        if getattr(model, "mask_head"):
+        if getattr(model, "mask_head", False):
             layers += [model.mask_head]
-        if getattr(model, "segm_head"):
+        if getattr(model, "segm_head", False):
             layers += [model.segm_head]
 
     elif isinstance(model, TwoStageDetector):
--- a/icevision/metrics/coco_metric/coco_metric.py
+++ b/icevision/metrics/coco_metric/coco_metric.py
@@ -58,10 +58,26 @@ class COCOMetric(Metric):
         self._reset()
         return logs
 
+    def metric_from_preds(
+        self,
+        preds: List[
+            Prediction
+        ],  # Prediction holds both the ground truth and the prediction records
+        print_summary: bool = False,
+    ) -> Dict[str, float]:
+
+        gt_list = [pred.ground_truth for pred in preds]
+        preds_list = [pred.pred for pred in preds]
+
+        logs = self.metric(
+            records=gt_list, preds=preds_list, print_summary=print_summary
+        )
+        return logs
+
     def metric(
         self,
-        records,
-        preds,
+        records,  # Ground Truth records
+        preds,  # Prediction records
         print_summary: bool = False,
     ) -> Dict[str, float]:
         with CaptureStdout():
--- a/icevision/metrics/coco_metric/coco_metric.py
+++ b/icevision/metrics/coco_metric/coco_metric.py
@@ -1,7 +1,6 @@
 __all__ = [
     "COCOMetric",
     "COCOMetricType",
-    "metric",
 ]
 
 from icevision.imports import *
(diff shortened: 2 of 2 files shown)


potipot:
I had such a use case, but cannot this be done with:
```python
inference_metric = COCOMetric(print_summary=True)
inference_metric.accumulate(preds)
inference_metric = coco_metric.finalize()
```

icevision/models/mmdet/utils.py
```diff
@@ -81,9 +81,9 @@ def param_groups(model):
         layers += [model.bbox_head]
 
         # YOLACT has mask_head and segm_head
-        if getattr(model, "mask_head"):
+        if getattr(model, "mask_head", False):
```
FraPochetti, line 84: why did we set this one to `False`?
same thing for line 86

FraPochetti:
> I had such a use case, but cannot this be done with:
> 
> ```python
> inference_metric = COCOMetric(print_summary=True)
> inference_metric.accumulate(preds)
> inference_metric_result = inference_metric.finalize()
> ```

Yeah, I guess it is the same. Just wrapping everything into a function call.

icevision/models/mmdet/utils.py
ai-fast-track: It's a bit confusing. I had a discussion with Dickson in his PR. I had the reaction as you.

Right now, there is a bug because False is missing

If False is missing it will add `mask_head` even for retinanet model for example.

The way to read is if `model` has a `mask_head` the statement is True otherwise set it to False.

ai-fast-track:
I didn't try that but I think it's better to have one simple and convenient method: No cognitive overload.

fstroth:
I think the proposal from Farid is good, I would just rename `metric_from_preds` to `evaluate_preds` so the following line 
`logs = inference_metric.metric_from_preds(preds, print_summary=True)`
would change to 
`logs = inference_metric.evaluate_preds(preds, print_summary=True)`
I think this is easier to remember and more clear.
Furthermore, I would rename the metric function as the name is very ambiguous and does not reflect what happens in the function, also I would make it a _ function as we are basically providing two ways to do the same thing. I would rename it to something like `_calculate_result`. Maybe we should also rename the `accumulate` and `finalize` to `_accumulate` and `_finalize` so we have a very clear API with an obvious way how to use it.
Any thoughts?
This might require quite some changes to multiple parts of the codebase, so I would offer to do all the changes as they are beyond the scope of this PR.

icevision/models/mmdet/utils.py
FraPochetti: Sorry man.
I still don't get it.
> If False is missing it will add mask_head
But that's what we want.

Here how I understand the following statement:
```python
if getattr(model, "mask_head"): layers += [model.mask_head]
```
if the `model` object has a "mask_head" attribute then add `model.mask_head` to the layers list.

What am I missing?

fstroth: The False is the default value of `getattr` if the object does not have the attribute, otherwise, an `AttributeError` will be raise if no default is provided. We should change `getattr` for `hasattr` as it is the function intended for the usecase.

FraPochetti:
I agree with @fstroth on everything he says.

Indeed a function named `metric` is extremely confusing.

As for @potipot 's point, why don't we add those 3 lines to the `getting_started` notebook just below Farid's solution?
I think it would show 2 different but very valid ways of achieving the same thing. A very powerful message.
```python
inference_metric = COCOMetric(print_summary=True)
inference_metric.accumulate(preds)
inference_metric_result = inference_metric.finalize()
```

icevision/models/mmdet/utils.py
FraPochetti: Ah, alright! Got it. Thanks Frederik.

fstroth:
There is some user interest in this PR. What do we need to do to get this merged?

PR #1061, opened 2022-02-08
Minor fix for clarity in mmdet utils.py

dnth:
This PR removes unwanted comments and minor fixes for code clarity.


FraPochetti:
Awesome stuff man! Thanks a lot!

dnth:
The test is failing.. is that a concern?

PR #1063, opened 2022-02-09
Support for DICOM images

hectorLop:
This PR adds support for DICOM images. I am preparing a notebook to show an example on [this](https://www.kaggle.com/c/rsna-pneumonia-detection-challenge/data?select=stage_2_train_labels.csv) dataset. I had to create the annotations.json because it is not provided. Also I have a few questions:
- Can icevision handle 16bit images? There are some medical images that are 16bit. Also, some albumentations transform cannot handle need the image to be in a 8bit format. I have already prepared the code to convert the image from 16bit to 8bit.
- Where should I add a new dependency? To handle the DICOM format I use the `pydicom` library.

### DIFF
--- a/icevision/imports.py
+++ b/icevision/imports.py
@@ -71,6 +71,8 @@ if SoftDependencies.wandb:
 if SoftDependencies.sklearn:
     import sklearn
 
+if SoftDependencies.pydicom:
+    import pydicom
 
 # TODO: Stop importing partial from fastcore and move this to utils
 class partial:
--- a/icevision/soft_dependencies.py
+++ b/icevision/soft_dependencies.py
@@ -27,6 +27,7 @@ class _SoftDependencies:
         self.sklearn = soft_import("sklearn")
         self.sahi = soft_import("sahi")
         self.fiftyone = soft_import("fiftyone")
+        self.pydicom = soft_import("pydicom")
 
     def check(self) -> Dict[str, bool]:
         return self.__dict__.copy()
--- a/icevision/utils/imageio.py
+++ b/icevision/utils/imageio.py
@@ -7,6 +7,7 @@ __all__ = [
     "plot_grid",
 ]
 
+import pydicom
 from icevision.imports import *
 from PIL import ExifTags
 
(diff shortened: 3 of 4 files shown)


fstroth:
We have an integration for 16bit images but it only supports tiff images atm. The Records are defined here: https://github.com/airctic/icevision/blob/master/icevision/core/record_defaults.py 
The only real difference is the GrayScaleRecordComponent defined here: https://github.com/airctic/icevision/blob/fabb086acdc2812c1ef1a882dcfcded814902521/icevision/core/record_components.py#L223

This RecordComponent loads the images with the `open_gray_scale_image` function, defined here: https://github.com/airctic/icevision/blob/fabb086acdc2812c1ef1a882dcfcded814902521/icevision/utils/imageio.py#L36

I think the best way to integrate Dicom images would be to add a check to the `open_gray_scale_image` function that checks if the image path ends on `.dmc` (we should use `.lower()` on the string) and use pydicom in that case. 
The conversion to 8bit should be separate augmentation I would say as in some cases one wants to keep the 16bit depth to preserve information. Furthermore, simple transformations like affine transformations don't require the images to be in 8bit.

FraPochetti:
This is epic!
Thanks a lot @hectorLop 
The best way to handle the new dependency is via the concept of `SoftDependency`.

Like [this](https://github.com/airctic/icevision/blob/fabb086acdc2812c1ef1a882dcfcded814902521/icevision/imports.py#L68) (example for `wandb`) and [here](https://github.com/airctic/icevision/blob/fabb086acdc2812c1ef1a882dcfcded814902521/icevision/soft_dependencies.py#L23)

hectorLop:
Thanks, both @fstroth and @FraPochetti for the guidelines! Great suggestions about implementing
this new feature into the codebase. I'll modify the PR to implement the changes!

hectorLop:
This is the [notebook](https://colab.research.google.com/drive/1EJkzR7ILeENMf2WKnMHWE5WeC2xAeL3Z?usp=sharing) to show that the implementation works. Note that the `kaggle download` command only downloads a few images rather than the full dataset, so it results in a training and validation sets of 16 and 4 images respectively. Therefore, the training is just done to show that the implementation works.

FraPochetti:
Thanks a ton @hectorLop for the contribution!

ai-fast-track:
Thanks a lot, @hectorLop!

PR #1084, opened 2022-04-01
don't open image on get_img_size

lgvaz:
fixes for normal images, but the problem still persists for dicom files


FraPochetti (changes requested):
Can you please make sure pre-commit hooks are running as expected?

PR #1086, opened 2022-04-05
[WIP] TorchvisionBackboneConfig refactor

dnth:
This PR proposes a refactor of `TorchvisionBackboneConfig` to adopt a similar structure to the `MMDetBackboneConfig`

Because `TorchvisionBackboneConfig` is used by existing Torchvision models such as `faster_rcnn`, `mask_rcnn`, `keypoint_rcnn` and `retinanet` modifying the `TorchvisionBackboneConfig` directly will cause the imports `from icevision.all import *` to fail. As a result, I could not get the getting started training notebook to run to test it out. 

Hence, I have create a temporary class `TorchvisionBackboneConfigV2` to work on the refactor. If the `TorchvisionBackboneConfigV2` passes all the code review by core-devs, this will replace all instances of `TorchvisionBackboneConfig` class. For the purpose of this test, I have also create a temporary `retinanetv2` model that uses the `TorchvisionBackboneConfigV2` class.

The model trains on my machine but with a relatively poor performance compared to the mmdet retinanet. The model I tested was instantiated with the following lines
```
model_type = models.torchvision.retinanetv2
backbone = model_type.backbones.resnet50_fpn
model = model_type.model(backbone=backbone(pretrained=True), num_classes=len(parser.class_map), **extra_args) 
```

image

Please let me know if this is consistent with the existing IceVision structure and how should I proceed further.
Thank you.


dnth:
@ai-fast-track do you think this is going in the right direction you intended? @lgvaz do you have any suggestions?

PR #1100, opened 2022-05-16
Remove manual `torch.no_grad` and train eval mode in LightningModule

aniketmaurya:
PL LightningModule automatically sets train and eval mode and `torch.no_grad` during the training and validation steps.

### DIFF
--- a/icevision/models/torchvision/lightning_model_adapter.py
+++ b/icevision/models/torchvision/lightning_model_adapter.py
@@ -36,17 +36,16 @@ class RCNNModelAdapter(LightningModelAdapter, ABC):
 
     def validation_step(self, batch, batch_idx):
         (xb, yb), records = batch
-        with torch.no_grad():
-            self.train()
-            train_preds = self(xb, yb)
-            loss = loss_fn(train_preds, yb)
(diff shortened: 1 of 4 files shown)


icevision/models/torchvision/lightning_model_adapter.py
```diff
@@ -36,17 +36,16 @@ def training_step(self, batch, batch_idx):
...
-
-            self.eval()
-            raw_preds = self(xb)
-            preds = self.convert_raw_predictions(
-                batch=batch, raw_preds=raw_preds, records=records
-            )
-            self.accumulate_metrics(preds=preds)
+        self.train()
+        train_preds = self(xb, yb)
+        loss = loss_fn(train_preds, yb)
```
aniketmaurya, line 41: I wasn't able to understand why do we calculate training prediction here?

potipot: Maybe the name is misleading , these are in fact validation preds. These are used to calculate validation loss and metrics

potipot:
Any idea why black formatting check might failing here?

Issue #1104, opened 2022-06-08
UnpicklingError: invalid load key, '<'.


GeorvityLabs:
When running in the Colab Notebook :
`model = model_type.model(backbone=backbone(pretrained=True), num_classes=len(parser.class_map), **extra_args) `

Getting the following error. Hope you may kindly suggest how to fix this!

```
load checkpoint from local path: checkpoints/vfnet/vfnet_r50_fpn_mstrain_2x_coco_20201027-7cc75bd2.pth
---------------------------------------------------------------------------
UnpicklingError                           Traceback (most recent call last)
[<ipython-input-10-696c7b7efa47>](https://localhost:8080/#) in <module>()
      1 # Instantiate the model
----> 2 model = model_type.model(backbone=backbone(pretrained=True), num_classes=len(parser.class_map), **extra_args)

7 frames
[/usr/local/lib/python3.7/dist-packages/torch/serialization.py](https://localhost:8080/#) in _legacy_load(f, map_location, pickle_module, **pickle_load_args)
    775             "functionality.")
    776 
--> 777     magic_number = pickle_module.load(f, **pickle_load_args)
    778     if magic_number != MAGIC_NUMBER:
    779         raise RuntimeError("Invalid magic number; corrupt file?")

UnpicklingError: invalid load key, '<'.
```

GeorvityLabs:
![Uploading issue_icevision.png…]()

GeorvityLabs:
![issue_icevision](https://user-images.githubusercontent.com/58555432/172707188-2bf3609e-dbfb-4b6e-ba2c-2c8cad72de3f.png)

jlvahldiek:
experienced the same issue with some mmdet models/backbones

GeorvityLabs:
@dnth any suggestions on how to fix this issue?

dnth:
Do you face this with other mmdet models? I think there are some changes to the mmdet models recently causing this issue.

I'd suggest trying out other mmdet models such as retinanet, or use efficientdet models.

dnth:
Hi @jlvahldiek we are in the process of fixing this on the IceVision repo.

If you're in a hurry, here is a Colab notebook as a workaround to make VFNet work again.

https://colab.research.google.com/gist/dnth/aea556c86ee3ca2e812ceb5eb8968caa/getting_started_object_detection.ipynb

dgleba:
This still seems to be an issue.

```
[...]

PR #1105, opened 2022-06-13
Update resnet_fpn.py

dnth:
This PR fixes the error `UnpicklingError: invalid load key, '<'` when loading a VFNet model.

Closes https://github.com/airctic/icevision/issues/1104


dnth:
Here is a Colab notebook to show VFNet works again with this PR.

https://colab.research.google.com/gist/dnth/aea556c86ee3ca2e812ceb5eb8968caa/getting_started_object_detection.ipynb

PR #1107, opened 2022-06-15
`TorchVisionBackbone` refactor

dnth:
This PR refactors the `TorchVisionBackbone` class. I've verified all models that uses the `TorchvisionBackbone` (UNet, Faster RCNN, Mask RCNN, Keypoint RCNN and Retinanet) to be working.

I tested out the torchvision retinanet in this notebook and now it is has better performance than before.

https://colab.research.google.com/gist/dnth/26275bdf9d79498046076618b8ff4a02/getting_started_object_detection.ipynb


dnth:
Here is the training epochs using the refactored Torchvision Retinanet

image

icevision/models/torchvision/fastai_callbacks.py
```diff
@@ -2,7 +2,8 @@
 
 from icevision.imports import *
 from icevision.engines.fastai import *
-from icevision.models.torchvision import faster_rcnn
+
+# from icevision.models.torchvision import faster_rcnn
```
fstroth, line 6: Will we need this line at a later point or can we just delete it?

fstroth (approved):
Looks good to me. Great work @dnth

icevision/models/torchvision/fastai_callbacks.py
dnth: Ahh good catch.. I will just remove it..

PR #1120, opened 2022-07-13
Lib restructure

fstroth:
His PR refactors ross and ultralytics to the new structure.


icevision/utils/model_creation.py
```diff
@@ -0,0 +1,232 @@
...
+    except AttributeError:
+        raise AttributeError(
+            f"Backbone: {backbone_name} not found. Possible options are: {get_backbone_names(backend_lib_name, model_type_name)}"
+        )
+
+    return model_type, backbone
+
+
+def build_model(
+    backend_lib_name: str,
```
ai-fast-track, line 192: Wdyt about simplifying the argument name by using:
`lib_name`, `model_name` instead of  `backend_lib_name` and  `model_type_name`

Also, would it be better to pass a dictionary instead of argument. Maybe using a dict, we will loose auto-completion.

PR #1123, opened 2022-07-18
update install script with lates mmdet dist to fix yolox input size i…

fabiogeraci:
I modified the install script in order to fix the input_size issue with yolox, as pointed out by @dnth

### DIFF
--- a/icevision_install.sh
+++ b/icevision_install.sh
@@ -45,7 +45,7 @@ esac
 
 
 echo "- Installing mmdet"
-pip install mmdet==2.17.0 --upgrade -q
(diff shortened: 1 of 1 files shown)


dnth:
Hi @fabiogeraci we've been wanting to upgrade the mmdet to 2.20.0 for a very long time now.. This change however breaks another model from mmdet (VFNet). Since there are many IceVision users who relied on VFNet we were reluctant to make this change. 

We are now planning for an upgrade that will include this and will be out very soon! Thank you again for submitting this PR.

fabiogeraci:
Thanks for the reply

PR #1127, opened 2022-07-30
Added directions on how to import IceVision without wildcard imports

### LINKED ISSUE #1116: How to use IceVision without using from icevision.all import *
## 📓 Documentation Update
**What part of documentation was unclear or wrong?**
To me, it is unclear which imports are needed in order to use ice vision.  
For instance, what is the import that is needed for the various models?  

**Describe the solution you'd like**
I'd like a clear description and examples of how to get started without using a wildcard import (`from icevision.all import *`).  

**Additional context**
In our case, we cannot use a wildcard import because our code is running as part of a [Kubeflow Pipelines](https://www.kubeflow.org/docs/components/pipelines/sdk-v2/v2-compatibility/).  
Long story short, it means our code is being executed inside a function, this function is then deployed to a docker image and executed. There is no such thing as a top level import for us.     
This means that we must use imports inside our main function, eg:  
Screenshot from 2022-07-04 14-55-47  
  
The following produces an error : `SyntaxError: import * only allowed at module level`

AlexandreBrown:
closes #1116

### DIFF
--- a/notebooks/getting_started_keypoint_detection.ipynb
+++ b/notebooks/getting_started_keypoint_detection.ipynb
@@ -118,6 +118,19 @@
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Alternative option of importing icevision\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


ai-fast-track:
Thanks a lot for your contribution, Alex!

PR #1130, opened 2022-08-10
Add a function to export preds to COCO annotations

boscacci:
I'm interested in using my partially-trained object detection model to help with the data labeling process. This new function helps facilitate that by enabling users to export inferences in COCO annotation format which align with the original images.


icevision/data/convert_records_to_coco_style.py
```diff
@@ -6,13 +6,104 @@
...
+        return json.JSONEncoder.default(self, obj)
+
+
+def export_batch_inferences_as_coco_annotations(
+    preds,
+    img_files,
+    transforms,
+    class_map,
+    output_filepath="new_pseudo_labels_for_further_training.json",
+):
```
ai-fast-track, line 44: It would be good to add `addl_info` an optional argument to let the user save that info in coco file

We could have: `info=None`  

We could also add `licenses` but maybe it's too much.

icevision/data/convert_records_to_coco_style.py
```diff
@@ -6,13 +6,104 @@
...
+    Returns
+    -------
+    None
+        This just spits out a serialized .json file and returns nothing.
+    """
+    object_category_list = [
+        {"id": v, "name": k, "supercategory": ""}
+        for k, v in class_map._class2id.items()
+    ]
+
```
ai-fast-track, line 74: We could check if `info` is None. If it isn't the case, we create the dict.

```
if info is None:
        "info": {
            "contributor": "",
            "date_created": "",
            "description": "",
            "url": "",
            "version": "",
            "year": "",
        }
else:
   info = info 

addl_info = {
        "licenses": [{"name": "", "id": 0, "url": ""}],
        "info": info,
        "categories": object_category_list,
    }

File: icevision/data/convert_records_to_coco_style.py [hunk 3]
```diff
@@ -6,13 +6,104 @@
...
+        The original array of fullres PIL images
+    transforms : list
+        list of transforms applied to original images (to be reversed)
+    class_map : icevision.core.class_map.ClassMap
+        The map of classes your model is training on
+    output_filepath : str, optional
+        The filepath (including filename) where you want the json results
+        to be serialized, by default
+        "new_pseudo_labels_for_further_training.json"
+
```
ai-fast-track, line 64:
```
   info : dict optional
     For example
     "info": {
        "year": "2022",
        "version": "1",
        "description": "Exported from IceVision",
        "contributor": "Awesome contributor",
        "url": "https://lazyannotator.fun",
        "date_created": "2022-08-05T20:13:09+00:00"
    },

icevision/data/convert_records_to_coco_style.py
boscacci: Good idea

PR #1131, opened 2022-08-11
Add dataset slicing and fix remove annotations

potipot:
# add dataset slicing
This simple PR adds functionality of slicing dataset to return a new sub dataset. This can be used in inference for example like so:
```
infer_dl = model_type.infer_dl(valid_ds[:4], batch_size=4, shuffle=False)
preds = model_type.predict_from_dl(model, infer_dl, keep_images=True)
```

# remove annotation
The remove annotation for detection component was broken, it only deleted the label_id from record but not the string encoded label itself.


ai-fast-track:
Great idea Pawel! Thanks!

potipot:
it seems like the error is coming from pytorch-lightning 1.7.0 
the version we install
```
Requirement already satisfied: pytorch-lightning>=1.4.5 in /opt/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages (from icevision==0.12.0) (1.7.1)
```
argument deprecated since 1.7.0:
https://pytorch-lightning.readthedocs.io/en/stable/generated/CHANGELOG.html#removed

PR #1132, opened 2022-08-11
Export preds to coco annotations

boscacci:
I'm interested in using my partially-trained object detection model to help with the data labeling process. This new function helps facilitate that by enabling users to export inferences in COCO annotation format which align with the original images.

This updated PR incorporates a suggestion to create an `addl_info` parameter in the function, in case the developer wants to create a more fleshed-out COCO annotation file with their name and license information. I also downsampled some of the new example images so they're not as big.


boscacci:
I thought `img_files` was a list of PIL images, but it's actually a list of file path references (type `fastcore.foundation.L`), so I've changed the docstring to reflect this. Later, PEP 484-style type hints should be added

PR #1133, opened 2022-08-12
Update tests to use newer pytorch lightning

potipot:
Parameters passed to pl.Trainer such a `weights_summary` and `checkpoint_callback` were deprecated since pl 1.5.0 and removed in 1.7.0. This PR updates test functions in order to support new pl verision.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,8 @@ inference =
 
 all =
     fastai >=2.5.2,<2.6
-    pytorch-lightning >=1.4.5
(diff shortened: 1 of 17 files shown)


potipot:
here we see that if we install the new pl version, the tests fail

potipot:
@ai-fast-track i'd consider it as high priority for merging cause due to this error our ci/cd pipeline is failing

fstroth (approved):
LGTM I think we can merge!

PR #1137, opened 2022-08-25
Clean up cluttered inference notebook output

boscacci:
Evidently "minimizing" notebook output is not the same as "clearing" it. This PR takes the "clearing" approach so that output is absent, rather than "minimized" (which still shows up in GitHub notebook rendering)


ai-fast-track:
Thanks a lot @boscacci!

PR #1147, opened 2022-11-01
Pin sahi version

boscacci:
Fix a breaking bug that was blocking `from icevision import *`

Approach: Set a max sahi version


AlexandreBrown (approved):

fcakyon:
Hello @boscacci @AlexandreBrown, I am the core maintainer of sahi. If you could describe to me where the sahi-related code is, I can update the source to support the latest version.

fcakyon:
@boscacci we can also update the min sahi version after upgrading to new API. Opened a related PR here: https://github.com/airctic/icevision/pull/1156

gatordevin:
Is there a reason this PR was abandoned? I understand the code can be fixed to work with the latest Sahi version but currently that is not the case and at the moment the default pip install is broken and requires pip installing a specific sahi version after the icevision install.

PR #1155, opened 2022-11-11
Updated import path for DetectionModel

Carlson-J:
sahi changed the import path to DetectionModel. This fixes that conflict.


fcakyon:
@Carlson-J thanks for the fix. 

It would be better to pin min sahi version to `0.11.1` since there have been many bug fixes and performance improvements since `sahi==0.8.*` 

I have opened a PR including these changes without knowing this one here: https://github.com/airctic/icevision/pull/1156

